PR #4, opened 2017-05-30
Added Defer Token

riginding:


riginding:
I've seen the 'finally' keyword in token.rs

whats the order of execution if there's finally and defer in the same function?

dinfuehr:
Thanks, great!

dinfuehr:
I would translate `defer` like this:

```
{
  <before>
  defer foo();
  <after>
}
```

should become this block:

```
{
  <before>;
  do {
    <after>
  } finally {
    foo();
  }
}
```

`finally` in `<before>`, `<after>` or sorrounding block shouldn't be important. Did you mean something else?

riginding:
what about?

``` dora
{
  <before>
  defer foo1();
  defer foo2();
  <after>
}
```
should this be:
``` dora
{
  <before>
  do {
    <after>
  } finally {
    foo2();
    foo1();
  }
}
```

riginding:
what about this? is this valid?

```
{
  do {
    <before1>
    defer foo();
    <after1>
  } finally {
    <before2>
    defer foo2();
    <after2>
  }
}
```

dinfuehr:
I wouldn't really try to optimize this in the baseline compiler, just generate nested blocks:

``` dora
{
  <before>
  do {
    do {
      <after>
    } finally {
      foo2();
    }
  } finally {
    foo1();
  }
}
```

The second example should also be valid and generated as this:

```
{
  do {
    <before1>
    do {
      <after1>
    } finally {
      foo();
    }
  } finally {
    <before2>
    do {
      <after2>
    } finally {
      foo2();
    }
  }
}
```

`defer` is just syntactic sugar for the first pattern. I think we can use these examples as tests...

PR #7, opened 2017-06-22
implement defer for typeck

riginding:


src/semck/typeck.rs
```diff
@@ -1795,6 +1805,19 @@ mod tests {
     }
 
     #[test]
+    fn type_defer() {
+        ok("fun foo() { }
+            fun f() { defer foo(); }") ;
+        err("fun foo(var i: int) { return i; }
```
dinfuehr: `UnknownIdentifier` is already thrown in `nameck` not in `typeck`. `typeck` isn't reached anymore, therefore we do not check what we thought we would. Sorry I didn't think about that...

Use a test like: `fun foo(a: int) {} fun f() { defer foo(); }` This should give the error `ParamTypesIncompatible` and is checked in typeck.rs

src/semck/typeck.rs
```diff
@@ -203,6 +203,16 @@ impl<'a, 'ast> TypeCheck<'a, 'ast> {
         }
     }
 
+    fn check_stmt_defer(&mut self, s: &'ast StmtDeferType) {
```
dinfuehr: `self.visit_expr(&s.expr)` is missing. This is necessary to type check the defer expression. Right now this method only checks if the given expression is a function call.

riginding:
@dinfuehr Implemented the changes.

riginding:
rebased on master

dinfuehr:
thanks!

PR #13, opened 2018-07-26
CLI functionality for create and manage projects, Imports and Includes and FFI 

playXE:
# Project Manager
For now i just added `start` function in lib.rs, so you can create rust project and add dora dependency and write something like this:
```rust
extern crate dora;

use dora::ctxt::SemContext

fn my_functions(ctxt: &mut SemContext) -> i32 {
      pub extern "C" fn add3(val: i32) -> i32 {
           return val + 3;
      }
      dora::native_fct(ctxt,"add3",add3 as *const u8);
      return 1;
}

fn main() {
     dora::start(std::path::Path::new("yourdorafile"),Some(my_functions as *const u8));
}
```
# W.I.P
1.  CLI tool for create and manage projects
2. Import keyword and functionality for it
3. Add more functionality for dora_ffi macro, like defining classes and etc
# Done

Include works, example: `include directory.myfile; 

Basic CLI functionality, `dora init <projectname>` creates rust project that have everything to work with dora
dora_ffi macro, it used to define functions that can be used in dora, example: 
```rust
dora_ffi! (ctxt;
    fun hello_world() {
         println!("Hello world!");
    }
)
```


dinfuehr:
Hi, nice work! If you would create a separate PR for removing LLVM would be great, easier to review and we could merge that pretty much immediately...

pigprogrammer:
@dinfuehr Okay, I'll create a pull request tomorrow, there's no time left today 😔

pigprogrammer:
@dinfuehr everything is ready to remove LLVM, i created another branch in my repo for CLI tool and other W.I.P thigs

pigprogrammer:
# Update

Include now works, you can write something like this:
```c
// file to include
fun test() {
    println("Include working!");
}
//main file
include myfile;; // for now you need to put two semicolons :(
fun main() {
     test() // prints "Include working!"
}

```

pigprogrammer:
Finally i started implementing `import` feature! :rocket:  But i need help, i don't know how to look for functions and etc in file to import, for now it looks like this: `import myfile:fun1,fun2;` or like this if you importing from directory: `import directory.and_file:fun1,fun3,fun4;`. @dinfuehr it would be good if you can help me make import working :+1:

pigprogrammer:
# Update
So i do some little upgrades and it seems like import closely to work, but there a problem, i don't know why but it looks for function with same name as import file:
```
error at 2:1: can not shadow function `toimport`.
```

dinfuehr:
Hi, first thanks for working on Dora! I hope I can do a good job reviewing and merging your changes. 

Usually this error comes up when a local variable shadows a function name, this is not allowed. It might be a bug or might be related that you added another type of names with imports/includes. I don't have good access to Dora for now, the error is probably thrown somewhere in the `semck` directory. Just grep for the error message and go from there.

Please come to me with any questions, I really want to help. Before you go on, could you please describe what you want to implement in a more specific way? Maybe describe the design you have in mind for include/import. You don't need to be too formal but we should somehow agree on a design first.

Please also keep in mind to keep PR's small, it's much faster for me to review that way. Thanks for your work!

pigprogrammer:
@dinfuehr 
> could you please describe what you want to implement in a more specific way?

Include is already implemented and it just used like C include to include everything from file, i tested it and it works perfectly! 
Import not implemented fully,how you can see there are some errors, i want to implement import like in dart language, in dart it looks like this: `import 'myfile.dart` `for function1,Vector2`, firstly i tried to implement it like this: `from myfile import function`, but it doesn't work fine so i just started implementing it like that: `import myfile:myfunction` and i get problems with Ast, i don't know it's good to create another ast for import file, and then merge funcs, for now it looks like this: 
```rust
                    ElemFunction(ref fun) => {
               
                        if interner.str(fun.name) == interner.str(name) {
                            // ast0 it's Ast that created in start() function
                            let mut edited_fun = fun.clone();
                            edited_fun.id = id_gen.next();
                            ast0.files[5].elements.push(ElemFunction(edited_fun));
                        } else {
                            return;
                        }
                    }
```

dinfuehr:
Hi, first thanks a lot for working on Dora! If we ever want to merge your code into Dora, you absolutely need to split your PR into many small self-contained PRs. This is much easier for me to review, such a big PR with multiple different features is impossible for me to review. For example you have a commit that adds REG_PARAMS for Windows. If you are working on Windows support you could create a PR just for this alone, it doesn't even have to be *full* Windows support but something that moves Dora into the right direction.

That's why I am closing this PR for now, just open small PRs for the individual features. This makes it easier for me to help as well.

PR #20, opened 2018-08-25
Fix #17: Swap arguments on floating point comparison to handle NaN correctly

### LINKED ISSUE #17: Incorrect floating point comparisons involving NaN
I believe all of these asserts should be correct, as all comparisons involving NaN should return `false`

```
    let nan = 0.0 / 0.0;
    assert(!(nan < 0.0)); // fail
    assert(!(nan > 0.0));
    assert(!(nan < nan)); // fail
    assert(!(nan > nan));
```

The first and the third fail though.

soc:
This change is hopefully in line with what you suggested!

I opted for some small code duplication, because I found it to be more readable, let me know if you like the change made differently!

### DIFF
--- a/src/masm/x64.rs
+++ b/src/masm/x64.rs
@@ -187,7 +187,7 @@ impl MacroAssembler {
                 asm::cmov(self, 0, dest, *scratch, CondCode::NotEqual);
             }
 
-            CondCode::Greater | CondCode::GreaterEq | CondCode::Less | CondCode::LessEq => {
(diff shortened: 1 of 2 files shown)


dinfuehr:
Looks great! IMHO code duplication is fine there. I don't expect that code to be changed that often and if then we can always merge those cases again anyways...

dinfuehr:
Thanks for fixing that!

PR #21, opened 2018-09-02
Update dependencies

soc:


soc:
The second commit allows updating `lazy_static` to 1.x, so I think this would be nice for the time being, even if it's planned to remove the LLVM dependency in the future.

dinfuehr:
I don't think the llvm part is necessary since I am going to remove it pretty soon...

PR #28, opened 2018-12-28
Boots - optimizing compiler for Dora

rodrigo-bruno:
This PR brings an initial version of the bytecode generator, which converts an AST into Bytecode.


dinfuehr:
Hi, finally had the chance to look into this. It seems that when you build the project in the test-configuration, the dependencies aren't included as `test` as well. Therefore that `parse`-function (and the `tests` module) can't be found. Since the function is quite short I guess we could just duplicate for the bytecode generator and keep `#cfg[test]`.

PR #29, opened 2019-03-16
Implement Int#asFloat, Long#asDouble, Float#asInt, Double#asLong ...

soc:
... which interpret the bits of the integer numbers as floating point values.

### DIFF
--- a/src/cpu/x64/asm.rs
+++ b/src/cpu/x64/asm.rs
@@ -1057,6 +1057,14 @@ pub fn cvttsd2si(buf: &mut MacroAssembler, x64: u8, dest: Reg, src: FReg) {
     sse_float_reg_freg(buf, true, 0x2c, x64, dest, src);
 }
 
+pub fn movd_f2i(buf: &mut MacroAssembler, x64: u8, dest: Reg, src: FReg) {
+    sse_movd_freg_reg(buf, 0x7e, x64, dest, src);
+}
+
+pub fn movd_i2f(buf: &mut MacroAssembler, x64: u8, dest: FReg, src: Reg) {
+    sse_movd_reg_freg(buf, 0x6e, x64, dest, src);
+}
+
 pub fn xorps(buf: &mut MacroAssembler, dest: FReg, src: Mem) {
     sse_float_freg_mem_66(buf, false, 0x57, dest, src);
 }
@@ -1126,6 +1134,30 @@ fn sse_float_reg_freg(buf: &mut MacroAssembler, dbl: bool, op: u8, x64: u8, dest
     emit_op(buf, 0x0f);
     emit_op(buf, op);
     emit_modrm(buf, 0b11, dest.and7(), src.and7());
+}
+
+fn sse_movd_reg_freg(buf: &mut MacroAssembler, op: u8, x64: u8, dest: FReg, src: Reg) {
+    emit_op(buf, 0x66);
+
+    if x64 != 0 || dest.msb() != 0 || src.msb() != 0 {
+        emit_rex(buf, x64, dest.msb(), 0, src.msb());
+    }
+
+    emit_op(buf, 0x0f);
+    emit_op(buf, op);
+    emit_modrm(buf, 0b11, dest.and7(), src.and7());
+}
+
+fn sse_movd_freg_reg(buf: &mut MacroAssembler, op: u8, x64: u8, dest: Reg, src: FReg) {
+    emit_op(buf, 0x66);
+
+    if x64 != 0 || dest.msb() != 0 || src.msb() != 0 {
+        emit_rex(buf, x64, src.msb(), 0, dest.msb());
+    }
+
+    emit_op(buf, 0x0f);
+    emit_op(buf, op);
+    emit_modrm(buf, 0b11, src.and7(), dest.and7());
 }
 
 pub fn pxor(buf: &mut MacroAssembler, dest: FReg, src: FReg) {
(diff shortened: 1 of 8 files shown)


soc:
This is work in progress, as running `target/debug/dora tests/float12.dora` results in:

```
error: trap not detected (signal 4, addr 0x7efbee5a222f).

ExecState {
    pc = 0x7efbee5a222f
    sp = 0x7ffe86862d20
    ra = 0x0
    regs[ 0] =                  0x3                    3
    regs[ 1] =       0x55a2dcc53e01       94157976976897
    regs[ 2] =                  0x1                    1
    regs[ 3] =       0x7ffe86863480      140731155362944
    regs[ 4] =       0x7ffe86862d20      140731155361056
    regs[ 5] =       0x7ffe86862d50      140731155361104
    regs[ 6] =       0x7efbee5a2180      139620500775296
    regs[ 7] =                  0x1                    1
    regs[ 8] =                  0x0                    0
    regs[ 9] =                  0x0                    0
    regs[10] =       0x7efbee5a2180      139620500775296
    regs[11] =       0x55a2dcb7adc0       94157976088000
    regs[12] =       0x7ffe86066000      140731146985472
    regs[13] =       0x7ffe86065000      140731146981376
    regs[14] =       0x7ffe868634b0      140731155362992
    regs[15] =       0x55a2dccbafb8       94157977399224
}

polling page = 0x7efbee5a1000

CodeMap {
  0x7efbee5a2000 - 0x7efbee5a207b => throw_thunk
  0x7efbee5a2080 - 0x7efbee5a2114 => trap_thunk
  0x7efbee5a2140 - 0x7efbee5a2290 => dora main()
  0x7efbee5a22c0 - 0x7efbee5a22e2 => dora_entry
```

Any hints welcome, as I'm lacking experience in both Rust and assembly. :-/

tests/float12.dora
```diff
@@ -4,4 +4,7 @@ fun main() {
 
     let x = 2.5D;
     assert(2.5F == x.toFloat());
+
+    let x = 3;
+    assert(4.2E-45F == x.asFloat());
 }
```
dinfuehr: We should add here a few more simple test cases, like `0`, `-0` or `NaN`. Those are values with well known bit representations. In both directions (float to int and int to float) and also for float and double.

src/cpu/x64/asm.rs
```diff
@@ -1057,6 +1057,10 @@ pub fn cvttsd2si(buf: &mut MacroAssembler, x64: u8, dest: Reg, src: FReg) {
     sse_float_reg_freg(buf, true, 0x2c, x64, dest, src);
 }
 
+pub fn movd(buf: &mut MacroAssembler, dest: FReg, x64: u8, src: Reg) {
```
dinfuehr: I usually add tests for assembly instructions at the bottom of the file. Please add a few test cases with different registers.  You can use [this online assembler](https://defuse.ca/online-x86-assembler.htm) to get the encoded bytes for an assembly instruction.

dinfuehr: I've run your PR locally and the failure is due to a SIGILL (an illegal instruction), so it seems to me that the instruction encoding is not right. Writing tests with `assert_emit!` combined with the online assembler should tell you what's wrong - not too much assembly required just run the unit tests until the instruction encodings match. If you need more help or don't want to look into this just ping me again.

src/masm/x64.rs
```diff
@@ -521,6 +521,28 @@ impl MacroAssembler {
...
     }
 
+    pub fn int_as_float(
+        &mut self,
+        dest_mode: MachineMode,
+        dest: FReg,
+        src_mode: MachineMode,
+        src: Reg,
+    ) {
+        asm::pxor(self, dest, dest);
```
dinfuehr: I suppose we don't need this all the time, clang for example doesn't emit this [here](https://gcc.godbolt.org/z/BtMbDb). I guess we could make it conditional like `if dest_mode.size() < src_mode.size() { asm::pxor(...) }`.

soc: Will do, thanks!

src/cpu/x64/asm.rs
soc: Thanks for the link, I'll have a look and try things the way you recommend. I'll report back.

src/masm/x64.rs
soc: I think we don't need it at all for the `as*` family of functions, as they only operate on src_mode.size() == dest_mope.size()`.

src/cpu/x64/asm.rs
soc: Your link was really helpful, thanks!

soc:
The asm tests fail with ...

```
failures:

---- cpu::x64::asm::tests::test_movd_f2i_32 stdout ----
exp: 66, 0f, 6e, c7
got: 66, 0f, 6e, f8
thread 'cpu::x64::asm::tests::test_movd_f2i_32' panicked at 'emitted code wrong.', src/cpu/x64/asm.rs:2128:9

---- cpu::x64::asm::tests::test_movd_f2i_64 stdout ----
exp: 66, 48, 0f, 6e, c7
got: 66, 48, 0f, 6e, f8
thread 'cpu::x64::asm::tests::test_movd_f2i_64' panicked at 'emitted code wrong.', src/cpu/x64/asm.rs:2133:9

---- cpu::x64::asm::tests::test_movd_i2f_32 stdout ----
exp: 66, 0f, 7e, c0
got: 66, 0f, 6e, c0
thread 'cpu::x64::asm::tests::test_movd_i2f_32' panicked at 'emitted code wrong.', src/cpu/x64/asm.rs:2138:9

---- cpu::x64::asm::tests::test_movd_i2f_64 stdout ----
exp: 66, 48, 0f, 7e, c0
got: 66, 48, 0f, 6e, c0
thread 'cpu::x64::asm::tests::test_movd_i2f_64' panicked at 'emitted code wrong.', src/cpu/x64/asm.rs:2143:9
```

... though I'm not quite sure how much of the failure is related to the non-availability of 32bit regs for testing in Dora.

I would appreciate any pointers!

src/cpu/x64/asm.rs
```diff
@@ -1057,6 +1057,14 @@ pub fn cvttsd2si(buf: &mut MacroAssembler, x64: u8, dest: Reg, src: FReg) {
     sse_float_reg_freg(buf, true, 0x2c, x64, dest, src);
 }
 
+pub fn movd_f2i(buf: &mut MacroAssembler, dest: Reg, x64: u8, src: FReg) {
+    sse_movd_freg_reg(buf, false, 0x6e, x64, dest, src);
```
dinfuehr: I think this should be 0x7e instead of 0x6e

src/cpu/x64/asm.rs
```diff
@@ -1128,6 +1136,30 @@ fn sse_float_reg_freg(buf: &mut MacroAssembler, dbl: bool, op: u8, x64: u8, dest
...
+    emit_op(buf, 0x0f);
+    emit_op(buf, op);
+    emit_modrm(buf, 0b11, dest.and7(), src.and7());
+}
+
+fn sse_movd_freg_reg(buf: &mut MacroAssembler, dbl: bool, op: u8, x64: u8, dest: Reg, src: FReg) {
+    emit_op(buf, 0x66);
+
+    if x64 != 0 || dest.msb() != 0 || src.msb() != 0 {
+        emit_rex(buf, x64, dest.msb(), 0, src.msb());
```
dinfuehr: swap src & dest here

src/cpu/x64/asm.rs
```diff
@@ -1128,6 +1136,30 @@ fn sse_float_reg_freg(buf: &mut MacroAssembler, dbl: bool, op: u8, x64: u8, dest
...
+    }
+
+    emit_op(buf, 0x0f);
+    emit_op(buf, op);
+    emit_modrm(buf, 0b11, dest.and7(), src.and7());
```
dinfuehr: swap src and dest here

src/cpu/x64/asm.rs
```diff
@@ -2091,6 +2123,26 @@ mod tests {
         assert_emit!(0xf2, 0x49, 0x0f, 0x2c, 0xe0; cvttsd2si(1, RSP, XMM8));
     }
 
+    #[test]
+    fn test_movd_f2i_32() {
+        assert_emit!(0x66, 0x0F, 0x6E, 0xC7; movd_f2i(RDI, 0, XMM0)); // EDI?
```
dinfuehr: 7e instead of 6e

src/cpu/x64/asm.rs
```diff
@@ -2091,6 +2123,26 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_movd_f2i_64() {
+        assert_emit!(0x66, 0x48, 0x0F, 0x6E, 0xC7; movd_f2i(RDI, 1, XMM0));
```
dinfuehr: 7e instead of 6e

src/cpu/x64/asm.rs
```diff
@@ -2091,6 +2123,26 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_movd_i2f_32() {
+        assert_emit!(0x66, 0x0F, 0x7E, 0xC0; movd_i2f(XMM0, RAX, 0)); // EAX?
```
dinfuehr: 6e instead of 7e

src/cpu/x64/asm.rs
```diff
@@ -2091,6 +2123,26 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_movd_i2f_64() {
+        assert_emit!(0x66, 0x48, 0x0F, 0x7E, 0xC0; movd_i2f(XMM0, RAX, 1));
```
dinfuehr: 6e instead of 7e

soc:
Thanks! Your help was completely on point, it was the exact issues I experienced. Now all the tests work, and I added a few additional roundtrip tests.

src/cpu/x64/asm.rs
```diff
@@ -2091,6 +2123,26 @@ mod tests {
         assert_emit!(0xf2, 0x49, 0x0f, 0x2c, 0xe0; cvttsd2si(1, RSP, XMM8));
     }
 
+    #[test]
+    fn test_movd_f2i_32() {
```
dinfuehr: I think we need only two test functions: `test_movd_f2i` and `test_movd_i2f`. Can you add a bit more tests with different registers, e.g. R8, R15, XMM7, XMM8, XMM15. Only like two or three combinations. It would be good to test them as well since those registers also set a bit in the rex-byte (register index is greater than 7 and therefore bit 3 is non 0)

src/masm/x64.rs
```diff
@@ -521,6 +521,44 @@ impl MacroAssembler {
...
     }
 
+    pub fn int_as_float(
+        &mut self,
+        dest_mode: MachineMode,
+        dest: FReg,
+        src_mode: MachineMode,
+        src: Reg,
+    ) {
+        let x64 = match src_mode {
```
dinfuehr: I think we should assert that `src_mode.size() == dest_mode.size()` here.

src/cpu/x64/asm.rs
```diff
@@ -1128,6 +1136,30 @@ fn sse_float_reg_freg(buf: &mut MacroAssembler, dbl: bool, op: u8, x64: u8, dest
...
+    if x64 != 0 || dest.msb() != 0 || src.msb() != 0 {
+        emit_rex(buf, x64, dest.msb(), 0, src.msb());
+    }
+
+    emit_op(buf, 0x0f);
+    emit_op(buf, op);
+    emit_modrm(buf, 0b11, dest.and7(), src.and7());
+}
+
+fn sse_movd_freg_reg(buf: &mut MacroAssembler, dbl: bool, op: u8, x64: u8, dest: Reg, src: FReg) {
```
dinfuehr: dbl is not needed as argument right? Could you also please make x64 the second argument here as well.

src/cpu/x64/asm.rs
```diff
@@ -1128,6 +1136,30 @@ fn sse_float_reg_freg(buf: &mut MacroAssembler, dbl: bool, op: u8, x64: u8, dest
     emit_modrm(buf, 0b11, dest.and7(), src.and7());
 }
 
+fn sse_movd_reg_freg(buf: &mut MacroAssembler, dbl: bool, op: u8, dest: FReg, x64: u8, src: Reg) {
```
dinfuehr: Same here

src/cpu/x64/asm.rs
```diff
@@ -1057,6 +1057,14 @@ pub fn cvttsd2si(buf: &mut MacroAssembler, x64: u8, dest: Reg, src: FReg) {
     sse_float_reg_freg(buf, true, 0x2c, x64, dest, src);
 }
 
+pub fn movd_f2i(buf: &mut MacroAssembler, dest: Reg, x64: u8, src: FReg) {
```
dinfuehr: x64 should be second argument here

src/cpu/x64/asm.rs
```diff
@@ -1057,6 +1057,14 @@ pub fn cvttsd2si(buf: &mut MacroAssembler, x64: u8, dest: Reg, src: FReg) {
...
+    sse_movd_freg_reg(buf, false, 0x7e, x64, dest, src);
+}
+
+pub fn movd_i2f(buf: &mut MacroAssembler, dest: FReg, src: Reg, x64: u8) {
```
dinfuehr: x64 should be second argument here

soc:
Sorry for the Cago.lock noise, that wasn't intentional. Should be gone now.

dinfuehr:
Thank you very much for the PR :+1:

soc:
@dinfuehr Thanks for all your help! :-D

Issue #24, opened 2018-10-19
Report file names on errors


soc:
One very important idea would be to report file names for errors.

In the past this wasn't necessary since there was only one file possible.

dinfuehr:
Here are some action points that could be done when working on this:

- [ ] Read source file into string and then parse/lex String instead of the File
- [ ] Keep all file strings in a table and use an index to identify a file
- [ ] Move file into `Position`
- [ ] `Position` then stores `file`, `line` and `column`. Let's have a location map and make `Position` an index into that map. `Position` should than be defined as `Position(u32)`, this should save some memory and allow for faster copying
- [ ] Keep indices of start of line numbers in a table, we can use this to print the full line in source code to highlight errors

soc:
@dinfuehr I think you did a lot of work here recently, do you feel this has been resolved for good?

soc:
Closing this, feel free to reopen, @dinfuehr!

PR #30, opened 2019-03-16
Add the file name when reporting compiler warnings/errors

soc:


soc:
This is a very rough sketch of tackling https://github.com/dinfuehr/dora/issues/24, does this go into the right direction? Creating `String`s all over the place is certainly not ideal ...

lib/dora-parser/src/error/msg.rs
```diff
@@ -369,22 +369,23 @@ impl Msg {
 
 #[derive(Clone, Debug)]
 pub struct MsgWithPos {
-    pub msg: Msg,
+    pub file: String,
```
dinfuehr: We should probably make this `Option<String>` or `Option<File>` if we introduce this vector of file names with this PR.

dinfuehr:
Hi, very cool! This definitely goes into the right direction! I especially like that you added a separate `report_with_file` function, such that we can introduce file names to error messages incrementally.

There are some tests failing with this PR, try `tools/test` locally to see which or follow the Travis link. They probably fail because these tests check the actual position of an error message with a regex in `def read_error_message` in `tools/tester.rb` and that regex isn't going to match `error1` or `error2`. I try to keep the tests succeeding at all times, we should solve this before merging this PR.

The cloning is definitely not ideal, I would propose to keep all parsed file names in an vector and to just use the index into a vector to identify a specific file with e.g. `pub struct File(u32);`. Depending on how much work this implies we can do it in this PR or in a subsequent one. What do you think about this?

soc:
Thanks for the help!

> There are some tests failing with this PR, try tools/test locally to see which or follow the Travis link. They probably fail because these tests check the actual position of an error message with a regex in def read_error_message in tools/tester.rb and that regex isn't going to match error1 or error2. I try to keep the tests succeeding at all times, we should solve this before merging this PR.

Yes, that's the case. Wanted to fix them after you were ok with the approach and the formatting/spelling of the message, so I don't have to touch it twice.

> The cloning is definitely not ideal, I would propose to keep all parsed file names in an vector and to just use the index into a vector to identify a specific file with e.g. pub struct File(u32);. Depending on how much work this implies we can do it in this PR or in a subsequent one. What do you think about this?

I'd definitely have a look; if you tell me where you want the vector to live (and how long) I can give it a shot.

dinfuehr:
I just noticed that we store all files in the [AST](https://github.com/dinfuehr/dora/blob/master/lib/dora-parser/src/ast/mod.rs#L15). Maybe we could just index into this vector.

lib/dora-parser/src/error/msg.rs
```diff
@@ -369,22 +369,23 @@ impl Msg {
...
     pub fn message(&self) -> String {
-        format!("error at {}: {}", self.pos, self.msg.message())
+        format!("error1 in {} at {}: {}", self.file, self.pos, self.msg.message())
     }
 }
 
 impl fmt::Display for MsgWithPos {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-        write!(f, "error at {}: {}", self.pos, self.msg.message())
+        write!(f, "error2 in {} at {}: {}", self.file, self.pos, self.msg.message())
```
dinfuehr: I think we should make this function something like:

```rust
if let Some(file) = self.file {
    write!(f, "error in {} at {}: {}", self.file, self.pos, self.msg.message())
} else {
    write!(f, "error at {}: {}", self.pos, self.msg.message())
}
```

In our Ruby script we can then just parse both error formats.

lib/dora-parser/src/error/msg.rs
dinfuehr: I am thinking whether we should move file into `Position`? I think this should probably be our ultimate goal but not sure whether this should (can) be done in this PR.

Please don't feel obliged to solve everything with this single PR. As long as you improve the current status quo it is good enough already (it already is).

FWIW I've once thought about keeping all files in memory for error messages [here](https://github.com/dinfuehr/dora/blob/master/lib/dora-parser/src/lexer/map.rs). But we should definitely not try to tackle that in this PR.

soc:
Just a heads up: I still agree on both your approach with storing an index into a vector and moving the path into `Position`.

I tried to get half-way there, addressing at least the latter in this commit. But adding the `String` to `Position` would have made `Position` not support `Copy`, and would have necessitated a lot of changes – all which would have to be undone after the move to index-into-vector.

I hope this is a gradual improvement that is still acceptable to you until I can work on moving things fully to the design you proposed.

dinfuehr:
It's definitely an improvement. The good thing about having the filename not in Position but only error messages is that we shouldn't have way more clones than before.

PR #31, opened 2019-03-17
Identity comparison (===) works on all types

soc:
- Already worked for references
- Now works for whole numbers (same as ==)
- Now works for floating point numbers such that
  float1  === float2  ----> float1.asInt()   == float2.asInt()
  double1 === double2 ----> double1.asLong() == double2.asLong()

### DIFF
--- a/src/baseline/expr.rs
+++ b/src/baseline/expr.rs
@@ -778,24 +778,7 @@ where
         if let Some(intrinsic) = self.intrinsic(e.id) {
             self.emit_intrinsic_bin(&e.lhs, &e.rhs, dest, intrinsic, Some(e.op));
         } else if e.op == BinOp::Cmp(CmpOp::Is) || e.op == BinOp::Cmp(CmpOp::IsNot) {
-            self.emit_expr(&e.lhs, REG_RESULT.into());
-            let offset = self.reserve_temp_for_node(&e.lhs);
-            self.asm
-                .store_mem(MachineMode::Ptr, Mem::Local(offset), REG_RESULT.into());
-
-            self.emit_expr(&e.rhs, REG_TMP1.into());
-            self.asm
-                .load_mem(MachineMode::Ptr, REG_RESULT.into(), Mem::Local(offset));
-
-            self.asm.cmp_reg(MachineMode::Ptr, REG_RESULT, REG_TMP1);
-
-            let op = match e.op {
-                BinOp::Cmp(CmpOp::Is) => CondCode::Equal,
-                _ => CondCode::NotEqual,
-            };
-
-            self.asm.set(dest.reg(), op);
-            self.free_temp_for_node(&e.lhs, offset);
+            self.emit_bin_is(e, dest.reg());
         } else if e.op == BinOp::Or {
             self.emit_bin_or(e, dest.reg());
         } else if e.op == BinOp::And {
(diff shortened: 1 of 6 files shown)


src/baseline/expr.rs
```diff
@@ -778,6 +778,19 @@ where
...
+            if mode.is_float() {
+                let intrinsic = match mode {
+                    MachineMode::Float32 => Intrinsic::FloatAsInt,
+                    MachineMode::Float64 => Intrinsic::DoubleAsLong,
+                    _ => unreachable!()
+
+                };
+                // expects &ExprCallType, e is Box<Expr>
+                //self.emit_intrinsic_float_as_int(e.lhs, REG_TMP1, intrinsic);
+                //self.emit_intrinsic_float_as_int(e.rhs, REG_TMP2, intrinsic);
```
soc: This is the place I'm stuck... help appreciated.

If I get floating-point values here, I want to convert them to integer values (with `movd`, as implemented earlier) and do a normal integer comparison on them.

dinfuehr: Don't go through an intrinsic here, you can directly MacroAssembler methods, I think it should look something like this pseudocode:

```
self.emit_expr(&e.lhs, FREG_RESULT.into()); // result is a float
mov_d2i(REG_RESULT, FREG_RESULT);
// save REG_RESULT on stack

self.emit_expr(&e.rhs, FREG_RESULT.into()); // result is a float
mov_d2i(REG_RESULT, FREG_RESULT);
// now compare REG_RESULT with value on stack
```

dinfuehr: Oh, forgot to mention: The non-float code path already stores a register on the stack and then compares both values. You can lookup the necessary code for these two comments above from there.

soc: Thanks, I'll check it out!

soc: Thanks for the help! :-)

soc:
Ah sorry, didn't see the other failing list.

src/baseline/expr.rs
```diff
@@ -778,14 +778,34 @@ where
...
+                self.asm.float_as_int(dest_mode, REG_RESULT, mode, FREG_RESULT);
+                offset = self.reserve_temp_for_node(&e.lhs);
+                self.asm
+                    .store_mem(MachineMode::Ptr, Mem::Local(offset), REG_RESULT.into());
 
-            self.emit_expr(&e.rhs, REG_TMP1.into());
-            self.asm
-                .load_mem(MachineMode::Ptr, REG_RESULT.into(), Mem::Local(offset));
+                self.emit_expr(&e.rhs, FREG_RESULT.into()); // result is a float
+                self.asm.float_as_int(dest_mode, REG_TMP1, mode, FREG_RESULT);
```
dinfuehr: I think this line is wrong, shouldn't it be REG_RESULT instead of REG_TMP1?

dinfuehr: Nevermind, not sure what I was thinking...

src/baseline/expr.rs
```diff
@@ -778,14 +778,34 @@ where
...
+                offset = self.reserve_temp_for_node(&e.lhs);
+                self.asm
+                    .store_mem(MachineMode::Ptr, Mem::Local(offset), REG_RESULT.into());
+
+                self.emit_expr(&e.rhs, REG_TMP1.into());
+                self.asm
+                    .load_mem(MachineMode::Ptr, REG_RESULT.into(), Mem::Local(offset));
+            }
 
             self.asm.cmp_reg(MachineMode::Ptr, REG_RESULT, REG_TMP1);
```
dinfuehr: But IIUC this line isn't correct anymore. `cmp_reg` uses always mode `Ptr` which was correct before since only references were compared. But now `===` also works for int, float or byte.

soc:
I fixed the failure in tests/nil2.dora

src/baseline/expr.rs
soc: Thanks, I updated the code!

It feels a bit clunky, but I couldn't find an existing function like this anywhere. Let me know if you want me to factor out the mode computation into some function.

src/baseline/expr.rs
```diff
@@ -778,16 +778,37 @@ where
         if let Some(intrinsic) = self.intrinsic(e.id) {
             self.emit_intrinsic_bin(&e.lhs, &e.rhs, dest, intrinsic, Some(e.op));
         } else if e.op == BinOp::Cmp(CmpOp::Is) || e.op == BinOp::Cmp(CmpOp::IsNot) {
```
dinfuehr: Could you please move this code-block into a new function `emit_bin_is`?

soc: Sure!

dinfuehr:
Hi, I would like to merge your PR but `tests/generic-array-has.dora` seems to fail (at least on travis). Other than that: very cool!

soc:
Sorry for that, should be fixed!

soc:
Thanks!

PR #35, opened 2019-03-21
Make syntax for defining funs more consistent with lets and vars ...

soc:
... by allowing a single expression as a function body:

    fun zero() -> int = 0 * (42 - 23);

### DIFF
--- a/tests/fn9.dora
+++ b/tests/fn9.dora
@@ -1,0 +1,11 @@
+fun main() -> int {
+  return myadd(1, 2) - 3;
+}
+
+fun myzero() -> int = 0 * (42 - 23);
+
+fun myfour() -> int = return 4;
+
+fun myadd(x: int, y: int) -> int = x + y + myfour() * myzero();
+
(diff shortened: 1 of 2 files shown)


tests/fn9.dora
```diff
@@ -0,0 +1,11 @@
...
+  return myadd(1, 2) - 3;
+}
+
+fun myzero() -> int = 0 * (42 - 23);
+
+fun myfour() -> int = return 4;
+
+fun myadd(x: int, y: int) -> int = x + y + myfour() * myzero();
+
+fun mythrow(x: int) -> int = throw "foo";
```
dinfuehr: Right now this is only checked at runtime, but a function that can throw an exception should be declared as `throws`. This might be a failure in the future.

dinfuehr:
LGTM, there is only that one small comment.

dinfuehr:
Thanks :+1:

soc:
It's a pleasure! Thanks!

PR #37, opened 2019-03-22
Rename internal

soc:
I believe it might be a more precise name and frees up internal as a
possible visibility modifier, if we ever need it.


soc:
Please note that this branch is currently sitting on the filename branch, so it would make sense to review the PR about filename first, and then I'll rebase this branch onto master.

dinfuehr:
I am not so sure about renaming internal to intrinsic. native (similar to Java) probably would make more sense here. intrinsic is something that has special compiler semantics. Not all `internal` functions are intrinsics though, e.g. print/println.

Just to be clear: after rebasing the first commit would be gone? So nothing to review in that commit?

soc:
`@native` works, let's pick this.

dinfuehr:
True, but I guess that native/intrinsic separation is more an internal compiler thing. While users can write functions in native code (hopefully in the future they really can ;)), they won't be able to add intrinsics.

I would say that `native` functions backed by a real native function should still be FctKind::Native, while `native` functions backed by an instrinsic should still be FctKind::Intrinsic/Builtin.

soc:
Ok, great! I will rework things that way. Thanks!

soc:
Maybe we keep _internal_ as the field name for the list of functions inside classes etc. which are special (= added by  _native_ (FctKind::Native) or _builtin_ (FctKind::Builtin) functions. And builtin functions are exposed on the language surface as `native`?

soc:
Welp, that was painful ... but hopefully this is now consistent. :-)

lib/dora-parser/src/ast/mod.rs
```diff
@@ -599,7 +599,7 @@ impl Modifier {
             Modifier::Open => "open",
             Modifier::Override => "override",
             Modifier::Final => "final",
-            Modifier::Internal => "internal",
+            Modifier::Builtin => "native",
```
dinfuehr: Don't you think we should name this `Modifier::Native` instead? All other modifiers have the same name as the keyword. I think this is a bit confusing when reading other code, e.g. in `parser.rs`.

src/ctxt.rs
```diff
@@ -892,7 +892,7 @@ impl FctKind {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub enum Intrinsic {
+pub enum Builtin {
```
dinfuehr: Actually I would have preferred `Intrinsic` to `Builtin` as a name, but if you prefer `Builtin` that's fine as well.

lib/dora-parser/src/parser.rs
```diff
@@ -289,7 +289,7 @@ impl<'a> Parser<'a> {
             name: ident,
             pos: pos,
             has_open: has_open,
-            internal: internal,
+            internal: builtin,
```
dinfuehr: When we don't use `internal` anymore, I think we should sooner or later also rename this field. But no need to do this in this commit.

lib/dora-parser/src/ast/mod.rs
soc: True. I was concerned that `Modifier::Native` being turned into `FctKind::Builtin` and `FctKind::Native` being something completely different would be somewhat confusing. Shall I rename it?

Or maybe we can find a different name for `FctKind::Native` and its "native" functions (that are unrelated to `@native`)?

lib/dora-parser/src/parser.rs
soc: +1. Seems like an oversight.

PR #38, opened 2019-03-23
Rename Str to String

soc:
It's more consistent with the toString function, and in general avoiding
abbreviations makes things slightly more approachable as there are often
multiple hypothetical abbreviations, but usually only one way to write
the full name.

My mid-term view is that I would like to prefer consistent, non-
abbreviated names everywhere, while allowing users to rename these names
to shorter ones as required on import.

All of this, of course, within reasons: The rule of thumb I propose for
abbreviations in Dora is this:

> If more people know the abbreviation than the full name,
> or if the abbreviations is used so overwhelmingly such
> that some people have issues coming up with the correct
> full name, then these abbreviations have become names of
> their own – they are fine to use.

Examples:
- Http instead of Hypertext Transfer Protocol
- Html instead of Hypertext Markup Language
- Css  instead of Cascading Style Sheets

### DIFF
--- a/bench/raytracer/primitive.dora
+++ b/bench/raytracer/primitive.dora
@@ -11,7 +11,7 @@ open abstract class Primitive {
 
     abstract fun normal(pnt: Vec3) -> Vec3;
     abstract fun intersect(ry: Ray) -> Isect;
-    abstract fun toString() -> Str;
+    abstract fun toString() -> String;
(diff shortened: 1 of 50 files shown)


dinfuehr:
:+1:

soc:
Thanks!

PR #40, opened 2019-03-24
Add Nothing type

soc:
This allows a more precise typing and a distinction between functions
that return no value of interest (Unit) and functions that never return
(Nothing), due to throwing exceptions, looping forever.

This allows making `throw ""` and `fun f() -> Nothing = throw ""; f()`
behave consistently from a typing perspective, which is required to
gracefully deal with e. g. if-expressions' definite assignment rules.

### DIFF
--- a/src/baseline/expr.rs
+++ b/src/baseline/expr.rs
@@ -1958,7 +1958,8 @@ fn result_reg(mode: MachineMode) -> ExprStore {
 fn check_for_nil(ty: BuiltinType) -> bool {
     match ty {
         BuiltinType::Error => panic!("error shouldn't occur in code generation."),
-        BuiltinType::Unit => false,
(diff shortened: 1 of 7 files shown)


dinfuehr:
Already LGTM :+1: We might want to wait until #37 landed and then rebase this PR. Is this the reason this PR is still marked as Draft or is there's something else in this PR missing that I haven't thought about?

soc:
> Is this the reason this PR is still marked as Draft or is there's something else in this PR missing that I haven't thought about?

Yes, one of the reason. :-) The other one is that it doesn't work properly yet: it probably needs similar treatment to `Unit` is some cases, but I haven't fully looked into it. (Fails with "no machine mode for nothing".)

Issue #42, opened 2019-03-24
Consistent naming for types


soc:
Currently types are written with an uppercase letter at the start, with the exception of primitive types.

I propose that all types should start with an uppercase letter for increased consistency.

This means aligning the naming of primitive types and renaming `int`, `long` etc. to `Int`, `Long` etc. A diverse set of languages ranging from Haskell, Scala, Kotlin, Ceylon and others have shown that this increased consistency comes without drawbacks.

Additionally, with special types like `Unit`, `Nothing` it's not immediately clear whether they should be considered primitive or reference types (let alone structs), so having a consistent naming scheme allows us to not worry about awkward casing choices and avoids requiring in elaborate language lawyering to explain the casing decision.

PR #45, opened 2019-03-24
Consistent naming for types

soc:
Before this, types were written with an uppercase letter at the start,
with the exception of primitive types.

This commit aligns the naming of primitive types and renames `int`,
`long` etc. to `Int`, `Long` etc.

A diverse set of languages ranging from Haskell, Scala, Kotlin, Ceylon
and others have shown that this increased consistency comes without
drawbacks.

Additionally, with special types like `Unit`, `Nothing` it would not
have been immediately clear whether they should be considered primitive
or reference types (let alone structs), so having a consistent naming
scheme allows us to not worry about awkward casing choices and avoids
requiring in elaborate language lawyering to explain the casing decision.


soc:
This resolves #42.

PR #50, opened 2019-03-28
Introduce Hash, Sortable, Default; start splitting prelude, other files

soc:


soc:
This is preparing some work on the standard library.

stdlib/Default.dora
```diff
@@ -0,0 +1,3 @@
+trait Default {
+  static fun default() -> Self;
```
soc: This should be `let`.

stdlib/Equals.dora
```diff
@@ -0,0 +1,7 @@
+trait Equals {
+  fun equals(other: Self) -> Bool;
```
soc: I think it would be better to not have the distinction between `equals` and `==`.

stdlib/Equals.dora
```diff
@@ -0,0 +1,7 @@
...
+}
+
+impl Equals for Int {
+  internal fun equals(other: Int) -> Bool;
```
soc: I think we should have a rule that a trait can be directly implemented by a class if it defines the function the trait requires. This would turn

```impl Equals for Int```

into a valid implementation.

PR #51, opened 2019-03-28
[WIP] Introduce modules, which work as both "packages" and "holders of static functions"

soc:


soc:
Detailed concept and design will follow, probably tomorrow.

PR #55, opened 2019-04-02
Generic syntax overhaul

soc:
Ok, this is probably the biggest change I ever want to make to Dora (and the one I feel is the most important one), so here is my reasoning:

Why `[]` instead of `<>` for generics?

- It makes function definitions and function calls look consistent:
```
// before
fun foo<T>(x: T) = ....
foo::<Int>(1);
// after
fun foo[T](x: T) = ....
foo[Int](1);
```
- Every language that ever tried to implement generics with `<>` had to deal with ugly workarounds:
  - In Rust there is `::<>` which people dislike to this day and want to fix.
  - C# does some elaborate hacks of parsing it as "less than" comparison, and if they figure out later that it isn't a comparison they go back and parse it as Generics.
  - It took C++ ages until it wouldn't try parsing `>>` as right shift (you had to use `> >`).
  - In Java you have really weird syntax to make parsing easier: `someInstance.<Integer>foo(1);`
  
- I'm very critical of Scala (which uses `[]`) but they got this absolutely right. To this day, no issues with this approach have been found, and I never heard of anyone wanting to go back to `<>`. So there is some practical evidence that this works as well as advertised.

- As a nice side-effect, it forces us to do away with special syntax for array access. It's not implemented in this PR, but I plan to bring array access back as `let x = someArray(1)` and `someArray(0) = 1`. This makes it extremely regular and treats array access simply as a function from some int to the array's element result type.

### DIFF
--- a/lib/dora-parser/src/parser.rs
+++ b/lib/dora-parser/src/parser.rs
@@ -324,9 +324,9 @@ impl<'a> Parser<'a> {
     }
 
     fn parse_type_params(&mut self) -> Result<Option<Vec<TypeParam>>, MsgWithPos> {
-        if self.token.is(TokenKind::Lt) {
+        if self.token.is(TokenKind::LBracket) {
             self.advance_token()?;
-            let params = self.parse_comma_list(TokenKind::Gt, |p| p.parse_type_param())?;
+            let params = self.parse_comma_list(TokenKind::RBracket, |p| p.parse_type_param())?;
 
             Ok(Some(params))
         } else {
@@ -561,7 +561,6 @@ impl<'a> Parser<'a> {
     fn parse_function(&mut self, modifiers: &Modifiers) -> Result<Function, MsgWithPos> {
         let pos = self.expect_token(TokenKind::Fun)?.position;
         let ident = self.expect_identifier()?;
-
         let type_params = self.parse_type_params()?;
         let params = self.parse_function_params()?;
         let throws = self.parse_throws()?;
@@ -724,9 +723,9 @@ impl<'a> Parser<'a> {
                 let pos = self.token.position;
                 let name = self.expect_identifier()?;
 
-                let params = if self.token.is(TokenKind::Lt) {
+                let params = if self.token.is(TokenKind::LBracket) {
                     self.advance_token()?;
-                    self.parse_comma_list(TokenKind::Gt, |p| Ok(Box::new(p.parse_type()?)))?
+                    self.parse_comma_list(TokenKind::RBracket, |p| Ok(Box::new(p.parse_type()?)))?
                 } else {
                     Vec::new()
                 };
@@ -1081,11 +1080,9 @@ impl<'a> Parser<'a> {
                     let tok = self.advance_token()?;
                     let ident = self.expect_identifier()?;
 
-                    let type_params = if self.token.is(TokenKind::Sep) {
+                    let type_params = if self.token.is(TokenKind::LBracket) {
                         self.advance_token()?;
-                        self.expect_token(TokenKind::Lt)?;
-
-                        Some(self.parse_comma_list(TokenKind::Gt, |p| p.parse_type())?)
+                        Some(self.parse_comma_list(TokenKind::RBracket, |p| p.parse_type())?)
                     } else {
                         None
                     };
@@ -1094,17 +1091,10 @@ impl<'a> Parser<'a> {
                         self.parse_call(tok.position, Some(left), Path::new(ident), type_params)?
 
                     } else {
+                        //println!("ident {:?}, pos {}", self.interner.str(ident), tok.position);
                         assert!(type_params.is_none());
                         Box::new(Expr::create_field(self.generate_id(), tok.position, left, ident))
                     }
-                }
-
-                TokenKind::LBracket => {
-                    let tok = self.advance_token()?;
-                    let index = self.parse_expression()?;
-                    self.expect_token(TokenKind::RBracket)?;
-
-                    Box::new(Expr::create_array(self.generate_id(), tok.position, left, index))
                 }
 
                 TokenKind::LParen => {
@@ -1193,17 +1183,15 @@ impl<'a> Parser<'a> {
         while self.token.is(TokenKind::Sep) {
             self.advance_token()?;
 
-            if self.token.is(TokenKind::Lt) {
-                self.expect_token(TokenKind::Lt)?;
+            let ident = self.expect_identifier()?;
+            path.push(ident);
+        }
 
-                let params = self.parse_comma_list(TokenKind::Gt, |p| p.parse_type())?;
-                type_params = Some(params);
-                break;
+        if self.token.is(TokenKind::LBracket) {
+            self.advance_token()?;
 
-            } else {
-                let ident = self.expect_identifier()?;
-                path.push(ident);
-            }
+            let params = self.parse_comma_list(TokenKind::RBracket, |p| p.parse_type())?;
+            type_params = Some(params);
         }
 
         // is this a function call?
@@ -2365,7 +2353,7 @@ mod tests {
 
     #[test]
     fn parse_type_basic_with_params() {
-        let (ty, interner) = parse_type("Foo<A, B>");
+        let (ty, interner) = parse_type("Foo[A, B]");
         let basic = ty.to_basic().unwrap();
 
         assert_eq!(2, basic.params.len());
@@ -2605,12 +2593,12 @@ mod tests {
         assert_eq!(2, call.args.len());
     }
 
-    #[test]
+    #[ignore] #[test]
     fn parse_array_index() {
-        let (expr, interner) = parse_expr("a[b]");
-        let expr = expr.to_array().unwrap();
-        assert_eq!("a", *interner.str(expr.object.to_ident().unwrap().name));
-        assert_eq!("b", *interner.str(expr.index.to_ident().unwrap().name));
+        let (expr, interner) = parse_expr("a(b)");
+        let expr = expr.to_call().unwrap();
+        assert_eq!("a", *interner.str(expr.clone().object.unwrap().to_ident().unwrap().name));
+        assert_eq!("b", *interner.str(expr.args[0].to_ident().unwrap().name));
     }
 
     #[test]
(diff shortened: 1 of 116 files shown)


soc:
Ok, this is probably the biggest change I want to make (and the one I feel is the most important one), so here is some reasoning:

Why `[]` instead of `<>` for generics?

- It makes function definitions and function calls look consistent:
```
// before
def foo<T>(x: T) = ....
foo::<Int>(1);
// after
def foo[T](x: T) = ....
foo[Int](1);
```
- Every language that ever tried to implement generics with `<>` had to deal with ugly workarounds:
  - In Rust there is `::<>` which people dislike to this day and want to fix.
  - C# does some elaborate hacks of parsing it as "less than" comparison, and if they figure out later that it isn't a comparison they go back and parse it as Generics.
  - It took C++ ages until it wouldn't try parsing `>>` as right shift (you had to use `> >`).
  - In Java you have really inconsistent syntax to make parsing easier: `int <T>foo(T x) { ... }; someInstance.<Integer>foo(1);`
  
I'm very critical of Scala (which uses `[]`) but they got this absolutely right. To this day, no issues with this approach have been found, and I never heard of anyone wanting to go back to `<>`.

- As a nice side-effect, it forces us to do away with special syntax for array access. It's not implemented in this PR, but I plan to bring array access back as `let x = someArray(1)` and `someArray(0) = 1`. This makes it extremely regular and treats array access simply as a function from some integer to the array's element result type.

dinfuehr:
Dora has the C++'s `> >` problem as well right now ;) I was too lazy to fix this.

Ok, but let's merge that PR when the new array syntax is ready.

lib/dora-parser/src/parser.rs
```diff
@@ -2591,9 +2578,9 @@ mod tests {
         assert_eq!(2, call.args.len());
     }
 
-    #[test]
+    #[ignore] #[test]
     fn parse_array_index() {
-        let (expr, interner) = parse_expr("a[b]");
+        let (expr, interner) = parse_expr("a(b)");
         let expr = expr.to_array().unwrap();
```
dinfuehr: I guess there will be no special array-syntax, this means you could probably already use the function call node here (`to_call`)

soc: Do you prefer this approach?

I assumed it would be better to parse `foo(bar)` as something like `Call(Select(foo), bar)` and figure out in the typechecker, whether this turns out to be a function call, or a selecting an array + get.

soc:
> Dora has the C++'s `> >` problem as well right now ;) I was too lazy to fix this.

:-)

> Ok, but let's merge that PR when the new array syntax is ready.

I would prefer doing it in two separate PRs, because I don't really have much insight regarding the changes necessary to do this. Would this be ok, too?

lib/dora-parser/src/parser.rs
dinfuehr: I think we are speaking of the same approach. What I meant is that `ExprArrayType` probably won't be necessary anymore since we already parse the new array syntax like `a(0)` as `ExprCallType`. That's why this test can just use `to_call` instead of `to_array`.

Yes, only the typechecker is able to detect whether this is an array access or a function call.

dinfuehr:
> I would prefer doing it in two separate PRs, because I don't really have much insight regarding the changes necessary to do this. Would this be ok, too?

In this case I would prefer doing it in one PR actually. It shouldn't be a big problem since I am not going to do major changes in the next few weeks, so rebasing should be trivial. Also I think that this doesn't require a lot of changes.

soc:
> In this case I would prefer doing it in one PR actually.

Ok, then I'll do it this way! If you can point me into the right direction, I would really appreciate it!

dinfuehr:
Sure, just let me know when you get stuck or need help.

soc:
Ok I now dropped the ExprArray stuff in https://github.com/dinfuehr/dora/pull/55/commits/f5e6bf1597155107cdd1a493f55b200d2d6b9d48, tests didn't change.

What should I do to address the `UnknownMethod` issue? I feel that simply adding more functions to `Array` in prelude.rs is not the right approach (and probably doesn't work anyway) ...

dinfuehr:
Ok, this seems to be a much harder than expected.

The problem ExprCallType is too unflexible right now. AFAIU it doesn't support `f(1,2,3)(3)` which is now valid when `f` has Array as its return value. I think it's a good time to fix that, this would also be necessary for lamdas.

I think the best way to continue is to create an additional PR (or multiple PRs if needed) that should change the underlying implementation of calls without changing the syntax. As soon as this PR gets in, we can merge this one and this PR will then be quite simple.

The goal for this new PR should be changing ExprCallType:
```rust
#[derive(Clone, Debug)]
pub struct ExprCallType {
    pub id: NodeId,
    pub pos: Position,

    pub path: Path, // REMOVE ME
    pub object: Option<Box<Expr>>, // REMOVE ME
    pub callee: Box<Expr>, // ADD ME
    pub args: Vec<Box<Expr>>,
    pub type_params: Option<Vec<Type>>,
}
```

The problem is that right now we can only write either `SomeClass::y()` or `<obj>.<name>()`. With the array syntax we need to be able to parse `<any expression>()`. Changing this will probably also require other changes first:

* Add `ExprPathType` as a new AST-node for storing <EXPR1>::<EXPR2>::... Apart from `id` and `pos` ExprPathType should store `lhs: Box<Expr>` and `rhs: Box<Expr>`

* Adapt `parse_primary` in parser.rs to also check for TokenKind::Sep and return ExprPath nodes in this case. Although in valid code this code path shouldn't be reached for now. Also add checks in semck that the separator is used on the right constructs. Right now this is probably just `Class::function_name` used for static methods.

* Adapt `parse_primary` to also check for TokenKind::LParen and return ExprCall for it with the expression returned from `parse_factor` used as `callee`. This code patch shouldn't be reached for valid code for now. Don't try removing `path` and `object` for now.

* Remove parsing of calls and separators in `parse_identifier_or_call` which then becomes `parse_identifier`.

* Remove parsing of calls for `dot` expressions in `parse_primary`. I think we should also parse TokenKind::Dot differently and allow all expressions after dot like this `<expr1>.<expr2>()`. Only in semck we should check whether `expr2` is a single identifier. This should help us in supporting LSP, where we most likely should be able to parse as much code as possible for good results.

* Only now try to remove `path` and `object`. This will require more changes in the typechecker and baseline JIT.

See this more as general guidelines, maybe it makes sense to implement this in a different order or some things are different than I expected them to be.

dinfuehr:
We can change the syntax but only at the very end. The internal changes required are much more important. Changing syntax should be quite simple then.

dinfuehr:
Actually instead of adding a separate ExprPathType node, we can just reuse ExprBinType and add Separator as a BinOp. Choose what makes sense to you.

soc:
@dinfuehr Thanks for the hints, do you have any advice on the `UnknownMethod` issue? I think without resolving this, I can't do the further steps you outlined.

dinfuehr:
If this is an error in the current PR then just ignore this failure for now. Without more information, I would suppose this error comes from the typechecker. As soon as we have AST nodes that can actually represent the new use cases (`<expr>()`), this should be much simpler to solve.

We can then just check whether the <expr> is an Array and then allow "function call" syntax on it.

soc:
I think the current error is correct, valid in a sense: It has some `arrayOfInts.set(1, "test")` and it fails because it cannot find a method `set(x: Int, y: String)`, because we only add `set(Int, T)`, but not `set(Int, String)` or any other variant.

dinfuehr:
I will try to take a look.

soc:
Thanks!

soc:
I believe there is a `specialize` missing somewhere for `Array`, that wasn't done before because the methods were never called directly, and instead the array accesses were intrinsified right-away.

soc:
Status:

Resolved both issues, currently looking into generating the correct code for array accesses in `emit_call`.

dinfuehr:
@soc: If you are okay with it, I could start working on this in a few days? Implement the necessary changes, such that we can merge this PR.

soc:
> If you are okay with it, I could start working on this in a few days? Implement the necessary changes, such that we can merge this PR.

@dinfuehr Absolutely! I didn't come around to it, because I was quite busy recently. :-/

soc:
Hey @dinfuehr, sorry for the long delay, I have been pretty busy recently!

I just rebased the PR against master, would you consider merging this PR without the `array.get(0)` -> `array(0)` implementation (of course after you are fine with the PR itself)?

Every time I wanted to work on the second part in the past, I first lost time to rebase the PR, and then it's after midnight again. :-/

src/semck/typeck.rs
```diff
@@ -4237,9 +4200,10 @@ mod tests {
...
     }
 
-    #[test]
+    #[ignore] #[test]
     fn test_assign_fct() {
         err(
+            // fixme: that gets me 2 errors: once on column 26, once on column 30
             "fun foo() {} fun bar() { foo = 1; }",
             pos(1, 30),
             Msg::FctReassigned,
```
soc: Not sure what to make of this ...

soc:
I cleaned up this PR – would this be ok this way?

soc:
I moved the to `x(y)` -> `x.get(y)` rewrite to https://github.com/dinfuehr/dora/issues/72.

soc:
Not sure why the CI is failing, is it because of the formatting?

dinfuehr:
I haven't looked into the failure but that could be the case. I've recently added rustfmt to the travis-tests.

dinfuehr:
I hope I can look into this tomorrow, otherwise it might take some time..

soc:
Your rustfmt commits seem to have made master green, I just pushed commits that should fix it for my PR, too.

stdlib/Array.dora
```diff
@@ -1,17 +1,17 @@
...
+  fun contains[T /*: Equals*/](value: T) -> Bool {
     var i = 0;
 
     while i < self.length() {
-      let x = self[i];
-      if /*x.equals(value) ||*/ x === value {
-        return true;
-      }
+      let x = self.get(i);
+      //if x.equals(value) || x === value { // TypesIncompatible("T", "T")
```
dinfuehr: If we can't express this function right now, we should probably remove it until we can.

soc: Ah, weird, I'm not even sure why this ended up in this commit... I'll fix it! Thanks!

PR #57, opened 2019-04-06
Add module keyword and allow parsing of module declarations

soc:
Everything else is a no-op right now.

### DIFF
--- a/lib/dora-parser/src/ast/mod.rs
+++ b/lib/dora-parser/src/ast/mod.rs
@@ -92,6 +92,7 @@ impl Ast {
 #[derive(Clone, Debug)]
 pub struct File {
     pub path: String,
+    pub module_decl: Option<ModuleDecl>,
     pub elements: Vec<Elem>,
 }
 
@@ -442,6 +443,13 @@ pub struct Class {
     pub methods: Vec<Function>,
     pub initializers: Vec<Box<Stmt>>,
     pub type_params: Option<Vec<TypeParam>>,
+}
+
+#[derive(Clone, Debug)]
+pub struct ModuleDecl {
+    pub id: NodeId,
+    pub pos: Position,
+    pub path: Vec<Name>,
 }
 
 #[derive(Clone, Debug)]
--- a/lib/dora-parser/src/parser.rs
+++ b/lib/dora-parser/src/parser.rs
@@ -61,6 +61,11 @@ impl<'a> Parser<'a> {
         self.init()?;
         let mut elements = vec![];
 
+        let mut module_decl = None;
+        if self.token.is(TokenKind::Module) {
+             module_decl = Some(self.parse_module_decl()?);
+        }
+
         while !self.token.is_eof() {
             self.parse_top_level_element(&mut elements)?;
         }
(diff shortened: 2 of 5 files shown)


lib/dora-parser/src/ast/mod.rs
```diff
@@ -92,6 +92,7 @@ impl Ast {
 #[derive(Clone, Debug)]
 pub struct File {
     pub path: String,
+    pub module_decl: Option<ModuleDecl>,
```
dinfuehr: Shouldn't this be one of the top-level elements?

lib/dora-parser/src/parser.rs
```diff
@@ -618,23 +638,33 @@ impl<'a> Parser<'a> {
                               mut parse: F)
                               -> Result<Vec<R>, MsgWithPos>
         where F: FnMut(&mut Parser) -> Result<R, MsgWithPos>
+    {
+        self.parse_list(stop, TokenKind::Comma, parse)
+    }
+
+    fn parse_list<F, R>(&mut self,
```
dinfuehr: Could you instead write a `parse_path` function that returns `Vec<Name>` that doesn't use `parse_list`? I think it would be better to not use parse_list. Right now parse_list can use the `;` terminator, but with closed modules like you proposed we would have a second separator `{`.

lib/dora-parser/src/parser.rs
```diff
@@ -61,6 +61,11 @@ impl<'a> Parser<'a> {
         self.init()?;
         let mut elements = vec![];
 
+        let mut module_decl = None;
```
dinfuehr: I think it is better to be more forgiving and parse this as top-level-element. We need to parse the open module `module x {}` as well anyways. In one of the checkers we can then report an error when there is more than one open module definition or it's not at the top of the file.

lib/dora-parser/src/parser.rs
soc: My intention was to cleanly separate modules-as-packages (open modules) and modules-as-holders-of-static-members (closed modules) from the get-go. Basically,

- an open module
  - can have multiple paths like `foo::bar::baz`
  - can only appear as the first element
  - needs to have its name start with a lowercase letter
  - ends only with `;`, it cannot have `{ ... }`

- a closed module
  - only ever has a single name, e. g. `Foo`, never a path with multiple elements
  - can appear everywhere
  - needs to have its name start with an uppercase letter
  - may end with `;`, or can contain members with `{ ... }`

I think there should never be a case where we can't decide upfront whether we are parsing an open or a closed method, so I'd just use a different `parse_...` method to parse closed modules.

dinfuehr: IIRC you mentioned LSP. I think we need parsing to be quite forgiving that this works well. E.g. open module might not be the first statement in a file. As soon as this is the case parsing would fail and not show any other errors.

You can still return either ModuleDecl or Module from `parse_module` depending on what form was used. Just because the AST can parse something like `module a::b {}` doesn't mean it's valid code, we can still report an error in semck.

soc: That's a good point, I think I will need to think for deeply about this. Thanks!

PR #58, opened 2019-04-06
Add implementations for Equals, Identity, Comparable, Sortable, Hash and Default

soc:


src/semck/typeck.rs
```diff
@@ -1827,6 +1827,7 @@ impl<'a, 'ast> MethodLookup<'a, 'ast> {
                 LookupKind::Ctor(cls_id) => {
                     let cls = self.ctxt.classes.idx(cls_id);
                     let cls = cls.read();
+                    println!("adding cstr! {:?}", self.ctxt.interner.str(cls.name));
```
dinfuehr: please remove this

stdlib/Bool.dora
```diff
@@ -1,5 +1,8 @@
 internal class Bool {
   internal fun toInt() -> Int;
+
+  fun hash() -> Int = self.toInt();
```
dinfuehr: Do we want `Long` as type for the hash?

tests/array15.dora
```diff
@@ -0,0 +1,53 @@
...
+
+    let x = arrayFill::<Int>(3, 3);
+    assert(arrayContains::<Int>(x, 3));
+    assert(arrayHas::<Int>(x, 3));
+
+    let x = arrayFill::<Long>(3, 3L);
+    assert(arrayContains::<Long>(x, 3L));
+    assert(arrayHas::<Long>(x, 3L));
+
+    // let x = arrayFill::<Double>(3, 0.0/0.0);
```
dinfuehr: I can look into this if you want, could you please open a bug with this?

src/semck/typeck.rs
soc: Oh damn, yes, that was a left-over from a different PR. Thanks!

stdlib/Bool.dora
soc: I have considered it, but I think if someone has problems with hashcode collisions with a space of 4 billion different values, another 32 bit won't save that person.

tests/array15.dora
soc: Yes, this would be great! I'm still very bad at debugging this register-related stuff.

soc: https://github.com/dinfuehr/dora/issues/59

dinfuehr: Yes, this is a bit messy in the compiler. We might have bytecode in the baseline compiler in the future, hopefully this will be a bit simpler than.

PR #61, opened 2019-04-06
Do not create publicly accessible constructors for class definitions without constructors

soc:


soc:
This is not totally correct – if a `class A` has field initializers, but no constructor, we probably still need something to call to initialize the fields in cases like `class B(...) : A` (or should we simply reject such class definitions altogether, @dinfuehr?).

dinfuehr:
Hi, sorry to look into this so late.... Please just ping me the next time this happens...

I don't think this works with field initializers. The constructor initializes these fields at the beginning, without the constructor this wouldn't happen. What's the intention behind the change? Getting rid of abstract?

soc:
Hi @dinfuehr,

don't worry!
I didn't ping you due to the problem you identified, which I didn't have time to sort out yet.

My motivation for this is that this feature where one didn't define a constructor, but got a public no-arg constructor "for free" always felt off in the languages that had it. The feature has this 1995-Java vibe to it, where everything is mutable (because otherwise the public no-arg would only create instances with the exact same data), and use of reflection for instance creation is pervasive.

 I'd prefer if Dora only had (publicly accessible¹) constructors if the user defined them.

¹ public accessible, because we need to generate _something_ to allow initialization of class A (that has no constructor) if a sub-class with a constructor extends class A.  Maybe having "protected" constructors could subsume this, but I haven't out the semantics yet.

PR #63, opened 2019-05-15
Added bytecode support for float literals

hzahradnik:


dinfuehr:
Nice work, thanks :+1:

PR #65, opened 2019-05-17
reference const string literals as index of string pool

hzahradnik:
string pool is currently not emitted into the bytecode.

### DIFF
--- a/src/bytecode/generate.rs
+++ b/src/bytecode/generate.rs
@@ -1,3 +1,4 @@
+use std::collections::hash_map::HashMap;
 use std::convert::From;
 use std::fmt;
 use std::mem;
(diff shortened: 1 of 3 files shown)


src/bytecode/generate.rs
```diff
@@ -783,12 +807,23 @@ impl BytecodeGenerator {
             .push(Bytecode::InvokeStaticPtr(dest, fid, start, num));
     }
 
+    fn generate_string_pool(&mut self) -> Vec<String> {
```
dinfuehr: I think this function should be defined as `fn generate_string_pool(map: HashMap<String, StrConstPoolIdx>) -> Vec<String>`

src/bytecode/generate.rs
```diff
@@ -783,12 +807,23 @@ impl BytecodeGenerator {
...
+        let mut pool : Vec<String> = vec![String::new(); self.string_pool_map.len()];
+        for (string_value, StrConstPoolIdx(index)) in self.string_pool_map.iter() {
```
dinfuehr: Here then just loop like this: `for ... in map`.

hzahradnik:
I adopted your suggestions, can you have a look?

PR #68, opened 2019-07-01
bytecode: added support for self expression

hzahradnik:


src/bytecode/astgen.rs
```diff
@@ -1410,7 +1437,7 @@ mod tests {
         let fct = code("fun f() -> Int { let x = 1; return x; }");
         let expected = vec![ConstInt(r(0), 1), RetInt(r(0))];
         assert_eq!(expected, fct.code());
-    }
+    }
```
dinfuehr: I think you added these spaces by mistake

src/bytecode/opcode.rs
```diff
@@ -32,6 +32,7 @@ pub enum Bytecode {
     MovFloat(Register, Register),
     MovDouble(Register, Register),
     MovPtr(Register, Register),
+    MovSelf(Register, Register),
```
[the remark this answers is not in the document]
hzahradnik: You're right. I will change it.

src/bytecode/astgen.rs
hzahradnik: Yep, I will remove them.

hzahradnik:
I updated the PR to reflect the special case of primitives (Bool, Byte, Int, ...). These types are implemented as classes, but self should not return a pointer, but the value itself.

dinfuehr:
Thanks!

PR #69, opened 2019-07-07
Added extern functions

playXE:
This PR adds external function declaration:
```kotlin
extern fun puts(str: Long) -> Int;

fun main() {
     puts(native_string("Hello,World!"));
}
```
# Why?
- No need to use `loadFunction`,`callN` functions
- No limitation to `Long` type in parameters.
- Easier than using `loadFunction` and `callN`


playXE:
I also think it would be nice to add `DLLlimport` keyword to do something like this:
```kotlin
extern DllImport ( "SDL2.dylib") fun SDL_Init(x Int) -> Int;
```
The current implementation of extern functions searches for function in libraries linked with a program so you can't just get access to function from SDL or another library.

playXE:
@dinfuehr I tried to use malloc/calloc to allocate some object, and this thing segfaults program, I think it will be good to mark objects allocated not by GC.

PR #70, opened 2019-08-03
cannon: init of new baseline compiler using byte code

hzahradnik:
I've started implementation of a new baseline compiler (called cannon), which uses the bytecode instead of the abstract syntax tree. Instead of visiting every node multiple times, the new compiler just generates the bytecode using the bytecode generator and then iterates over the bytecode list.

You can use the compiler with the command line flag `--bc=cannon`. With `--emit-bytecode=<fct>` cannon will dump the bytecode of the function before generating the assembler code. Currently cannon supports only assignment of constant boolean and return without a return value. E.g. `fun main() { let x = true: }` should be executable.

I also extended the test tools to support the testing of cannon:

- The test file modifier `//= cannon` is available to mark a test file to be executed with cannon
- With the command line argument `-bc=<standard|cannon>` it is possible to filter the test files to run only files with the given baseline compiler.

Maybe it would be better to drop the test file modifier, because if we switch permanently to the new  baseline compiler, we would have to mark every new test file to run with cannon, instead of being the default.


src/baseline.rs
```diff
@@ -1,6 +1,8 @@
 pub use crate::baseline::codegen::{generate, generate_fct};
 
 mod asm;
+pub mod ast;
```
dinfuehr: Shouldn't we also move `baseline/expr.rs` into the ast-module?

hzahradnik: Yes, you're right, that would be possible.
When I think about it, the remaining content of src/baseline/info.rs could also be moved to the src/baseline/ast/info.rs, as the JitInfo is only used by the AstCodeGen.

src/driver/cmd.rs
```diff
@@ -219,6 +226,12 @@ pub enum CollectorName {
     SweepSwiper,
 }
 
+#[derive(Copy, Clone, Debug, RustcDecodable)]
+pub enum BaselineName {
+    Cannon,
+    Standard,
```
dinfuehr: We should find a name for the old baseline compiler that we can keep even when this compiler isn't the default one anymore.

tools/tester.rb
```diff
@@ -19,15 +20,21 @@
   end
 end
 
+$ARGS.delete_if do |arg|
```
dinfuehr: Ok for now but we should get proper argument processing sooner or later.

tools/tester.rb
```diff
@@ -19,15 +20,21 @@
...
+  if (m = /\A\-bc=(standard|cannon)/.match(arg))
+    $filter_run = m[1]
```
dinfuehr: Can we name this `filter_config` or something? Do we even need this flag?

tools/tester.rb
```diff
@@ -36,6 +43,116 @@ def initialize(opts = {})
...
+    self.args = self.vm_args = ""
+    self.target = $release ? "release" : "debug"
+  end
+
+  def run()
+    if self.expectation == :ignore
+      self.results = :ignore 
+      return {:ignore => 1}
+    end
+    self.optional_runs.each_pair { |optional_run, run_vm_args|
```
dinfuehr: Use `do |optional_run, ..| ... end` instead of braces here.

tools/tester.rb
```diff
@@ -274,8 +336,8 @@ def read_cmdline(text)
   args
 end
 
-def test_case_expectation(file)
-  exp = TestExpectation.new(fail: false)
+def parse_test_file(file)
+  test_case = TestCase.new(file)
```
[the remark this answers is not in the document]
hzahradnik: I thought about that too, but didn't want to spend too much time on this.

tools/tester.rb
hzahradnik: We do not need it, if we keep the `// cannon` modifier inside the test cases. I added this to allow running all tests with a specific baseline compiler without modifiying each file. I will remove it.

src/baseline/cannon/codegen.rs
```diff
@@ -0,0 +1,152 @@
...
+        let polling_page = self.vm.polling_page.addr();
+        self.asm
+            .epilog_with_polling(bytecode.stacksize(), polling_page);
+    }
+
+    fn emit_const_bool(&mut self, bytecode: &BytecodeFunction, dest: &Register, bool_const: bool) {
+        let index = match *dest {
+            Register(index) => index,
+        };
+        let bytecode_type = match bytecode.registers().get(index) {
```
dinfuehr: You can use `unwrap()` or `expect()` here, should be a much more concise than `match` here.

src/baseline/cannon/codegen.rs
```diff
@@ -0,0 +1,152 @@
...
+            Some(bytecode_type) => bytecode_type,
+            _ => panic!("register not found"),
+        };
+        let offset = match bytecode.offset().get(index) {
```
dinfuehr: Same here

src/driver/cmd.rs
hzahradnik: I don't have any good idea for a themed name. In the code the "standard" baseline compiler is simply called AstCodeGen. We could that name everywhere.

src/baseline/cannon/codegen.rs
```diff
@@ -0,0 +1,144 @@
...
+    fn emit_epilog(&mut self, bytecode: &BytecodeFunction) {
+        self.asm.emit_comment(Comment::Newline);
+        self.asm.emit_comment(Comment::Lit("epilog"));
+
+        let polling_page = self.vm.polling_page.addr();
+        self.asm
+            .epilog_with_polling(bytecode.stacksize(), polling_page);
+    }
+
+    fn emit_const_bool(&mut self, bytecode: &BytecodeFunction, dest: &Register, bool_const: bool) {
```
dinfuehr: I think you can pass `dest` as value (just as `Register`)

src/baseline/cannon/codegen.rs
```diff
@@ -0,0 +1,144 @@
...
+        if should_emit_debug(self.vm, self.fct) {
+            self.asm.debug();
+        }
+
+        self.emit_prolog(&bytecode);
+        // self.store_register_params_on_stack();
+        for btcode in bytecode.code() {
+            match btcode {
+                Bytecode::ConstTrue(dest) => self.emit_const_bool(&bytecode, &dest, true),
+                Bytecode::ConstFalse(dest) => self.emit_const_bool(&bytecode, &dest, true),
```
dinfuehr: shouldn't you pass here `false`?

src/baseline/cannon/codegen.rs
```diff
@@ -0,0 +1,144 @@
...
+        if should_emit_bytecode(self.vm, self.fct) {
+            bytecode.dump();
+        }
+
+        if should_emit_debug(self.vm, self.fct) {
+            self.asm.debug();
+        }
+
+        self.emit_prolog(&bytecode);
+        // self.store_register_params_on_stack();
```
dinfuehr: please remove commented out code

src/baseline/cannon/codegen.rs
```diff
@@ -0,0 +1,144 @@
...
+        self.asm.emit_comment(Comment::Lit("epilog"));
+
+        let polling_page = self.vm.polling_page.addr();
+        self.asm
+            .epilog_with_polling(bytecode.stacksize(), polling_page);
+    }
+
+    fn emit_const_bool(&mut self, bytecode: &BytecodeFunction, dest: &Register, bool_const: bool) {
+        let Register(index) = *dest;
+        let bytecode_type = bytecode.registers().get(index).expect("register not found");
```
dinfuehr: AFAICS `index` is an integer here, couldn't we enforce `Register` here as index type?

src/baseline/cannon/codegen.rs
```diff
@@ -0,0 +1,144 @@
...
+                Bytecode::ConstTrue(dest) => self.emit_const_bool(&bytecode, &dest, true),
+                Bytecode::ConstFalse(dest) => self.emit_const_bool(&bytecode, &dest, true),
+                Bytecode::RetVoid => {}
+                _ => panic!("bytecode not implemented"),
+            }
+        }
+
+        let always_returns = self.src.always_returns;
+
+        if !always_returns {
```
dinfuehr: We shouldn't use always_returns in the new baseline compiler. The generated bytecode should always contain a `ret` at the end of the function.

src/baseline/cannon/codegen.rs
```diff
@@ -0,0 +1,139 @@
...
+            self.asm.load_false(REG_RESULT);
+        }
+        self.asm
+            .store_mem(bytecode_type.mode(), Mem::Local(*offset), REG_RESULT.into());
+    }
+}
+
+impl<'a, 'ast> CodeGen<'ast> for CannonCodeGen<'a, 'ast> {
+    fn generate(mut self) -> JitBaselineFct {
+        let bytecode = generate_fct(
```
dinfuehr: I think it makes sense to move the bytecode generation out of CannonCodeGen. Its responsibility is to generate Machinecode from the bytecode, but not to generate the bytecode. Then you could add the bytecode as a field in CannonCodeGen and you don't need to pass it to each function as argument.

src/baseline/cannon/codegen.rs
hzahradnik: I can not follow completely. What do you mean?

src/baseline/cannon/codegen.rs
hzahradnik: I agree. What do you think about a factory or something similar for the code generators? Then we could move the creation of the baseline compiler away from `baseline/codegen.rs` and reduce the boiler plate code. 
I would do this in another PR, so we could merge this on. Obviously, only if this is OK for you.

tools/tester.rb
```diff
@@ -36,6 +43,116 @@ def initialize(opts = {})
...
+                :args,
+                :target,
+                :expectation,
+                :optional_runs,
+                :results
+
+  def initialize(file, opts = {})
+    self.expectation = opts.fetch(:expectation, TestExpectation.new(fail: false))
+    self.file = self.test_file = file
+    self.optional_runs = {nil => ''}
```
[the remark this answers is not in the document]
hzahradnik: I guess this is a problem for its own PR. I will resolve this conversation for now.

PR #76, opened 2019-08-11
Add trait Stringable and class Option

soc:
This adds a `Stringable` trait (maybe I find a better name in the future :-)) and an implementation of an `Option` class.¹

¹ As far as I understand, we do not have support for lambdas yet?

### DIFF
--- a/stdlib/Option.dora
+++ b/stdlib/Option.dora
@@ -1,0 +1,28 @@
+// FIXME: Identity/Equals/Stringable constraints should all end up on the individual methods that need them, not on the class
+// After nil is gone, this class needs some more intrinsics, so that it can treat <nullptr> as an empty Option (and get rid
+// of the isPresent field). Usual caveats apply: this only works for the first "layer", not for Options inside Options, etc.
+class Option[T : Identity + Equals + Stringable](let value: T, let isPresent: Bool) {
+
+  fun isEmpty() -> Bool = !self.isPresent;
+
+  fun has(rhs: T) -> Bool = self.isPresent && self.value === rhs;
+
+  fun contains(rhs: T) -> Bool =
+    self.isPresent && (self.value === rhs || self.value.equals(rhs));
+
+  fun equals(rhs: Option[T]) -> Bool =
+    self.isPresent == rhs.isPresent && self.value.equals(rhs.value);
+
+  fun toString() -> String {
+    if self.isPresent {
+      return "Some(" + self.value.toString() + ")";
+    } else {
+      return "None";
+    }
+  }
+
+}
+
+fun none[T : Identity + Equals + Stringable]() -> Option[T] = Option[T](defaultValue[T](), false);
+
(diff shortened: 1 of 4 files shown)


stdlib/Option.dora
```diff
@@ -0,0 +1,30 @@
+// FIXME: Identity/Equals/Stringable constraints should all end up on the individual methods that need them, not on the class
+// After nil is gone, this class needs some more intrinsics, so that it can treat <nullptr> as an empty Option (and get rid
+// of the isPresent field). Usual caveats apply: this only works for the first "layer", not for Options inside Options, etc.
+class Option[T : Identity + Equals + Stringable](let value: T, let isPresent: Bool) {
```
dinfuehr: Do you think we should implement this using subclassing? This should save some memory.

stdlib/Option.dora
```diff
@@ -0,0 +1,30 @@
...
+      return "Some(" + self.value.toString() + ")";
+    } else {
+      return "None";
+    }
+  }
+
+}
+
+// error: no ctor with definition `Option(nil, Bool)`
+// fun none[T : Identity + Equals + Stringable]() -> Option[T] = Option[T](nil, false);
```
dinfuehr: While this is a bug or at least the error message super misleading, nil isn't compatible with every type. With subclassing we wouldn't have this problem...

stdlib/Option.dora
soc: Considering that I want to get rid of `nil`, I think I wouldn't bother with it. :-)

stdlib/Option.dora
soc: Dito.

soc:
Do you have an idea how far we are away from value types?

I think they are crucial to make `Option` really useful and a prerequisite to a lot of other things.

stdlib/Option.dora
dinfuehr: Completely forgot about that: for these cases there is `defaultValue()`.

dinfuehr:
I guess long enough, that changing this PR should pay off. The changes shouldn't be too dramatic. There is quite some work left until structs, ADTs and pattern matching work well enough.

stdlib/Option.dora
soc: Good idea! Done.

soc:
I updated the PR to use `defaultValue`, did you have any other changes in mind?

tests/stringable.dora
```diff
@@ -0,0 +1,13 @@
+fun main() {
+  str[Bool](true);
+  str[Bool](false);
+  str[Byte](1.toByte());
+  str[Int](1);
+  str[Long](1L);
+  // thread 'main' panicked at 'fp-register accessed as gp-register.'
```
dinfuehr: Is this still a bug here?

soc: No, shouldn't be, I'll update the commit.

Issue #75, opened 2019-08-11
true.not(): thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0'


soc:
This should be the same as `!true` (which works).

PR #77, opened 2019-08-12
Fix #74, #75: Implement {Int,Long}#{not}, {Int, Long, Float, Double}#{unaryPlus,unaryMinus}, Boolean#not

### LINKED ISSUE #74: 1.not(): thread 'main' panicked at 'unknown intrinsic IntNot'
I assume this should be the same as `!1` (which works)?

soc:


soc:
This is duplicating things a bit with `emit_un`, but I couldn't make calling that function instead work:

```
                Intrinsic::IntNeg => {
                    let opnd = e.object.as_ref().unwrap().clone();
                    let unOp = ExprUnType { id: e.id, pos: e.pos, op: UnOp::Neg, opnd };
                    self.emit_un(&unOp, dest)
                },
                Intrinsic::IntNot => {
                    let opnd = e.object.as_ref().unwrap().clone();
                    let unOp = ExprUnType { id: e.id, pos: e.pos, op: UnOp::Not, opnd };
                    self.emit_un(&unOp, dest)
                },
```

doesn't work due to:

```
error[E0597]: `unOp` does not live long enough
    --> src/baseline/expr.rs:1046:34
     |
94   | impl<'a, 'b, 'ast> ExprGen<'a, 'b, 'ast>
     |              ---- lifetime `'ast` defined here
...
1046 |                     self.emit_un(&unOp, dest)
     |                     -------------^^^^^-------
     |                     |            |
     |                     |            borrowed value does not live long enough
     |                     argument requires that `unOp` is borrowed for `'ast`
1047 |                 },
     |                 - `unOp` dropped here while still borrowed
```

Let me know which way you prefer!

src/baseline/expr.rs
```diff
@@ -1392,13 +1396,49 @@ where
...
 
         self.emit_intrinsic_bin(lhs, rhs, dest, intr, None);
     }
 
+    fn emit_intrinsic_unary(
+        &mut self,
+        lhs: &'ast Expr,
+        dest: ExprStore,
+        intr: Intrinsic,
+        op: Option<BinOp>,
```
dinfuehr: `op` isn't used here, also pretty sure the type is wrong. For unary operators it should be `UnOp`.

src/baseline/expr.rs
```diff
@@ -1029,7 +1029,7 @@ where
                 Intrinsic::ByteNot => self.emit_intrinsic_bin_call(e, dest, intrinsic),
 
                 Intrinsic::BoolEq => self.emit_intrinsic_bin_call(e, dest, intrinsic),
-                Intrinsic::BoolNot => self.emit_intrinsic_bin_call(e, dest, intrinsic),
+                Intrinsic::BoolNot => self.emit_intrinsic_unary_call(e, dest, intrinsic),
```
dinfuehr: I think it's better to have a separate function for unary intrinsics.

src/baseline/expr.rs
soc: Will do!

src/baseline/expr.rs
soc: Will do!

soc:
@dinfuehr Does this look good?

src/baseline/ast/expr.rs
```diff
@@ -1443,10 +1467,12 @@ where
             Intrinsic::IntMul => self.asm.int_mul(MachineMode::Int32, dest, lhs, rhs),
             Intrinsic::IntDiv => self.asm.int_div(MachineMode::Int32, dest, lhs, rhs),
             Intrinsic::IntMod => self.asm.int_mod(MachineMode::Int32, dest, lhs, rhs),
+            Intrinsic::IntNeg => self.asm.int_neg(MachineMode::Int32, dest, lhs),
```
dinfuehr: Can you please move these unary intrinsics out of `emit_intrinsic_bin`? `lhs`, `rhs` and `BinOp` doesn't make sense for them.

src/baseline/ast/expr.rs
```diff
@@ -992,10 +1006,13 @@ where
                 Intrinsic::IntMul => self.emit_intrinsic_bin_call(e, dest, intrinsic),
                 Intrinsic::IntDiv => self.emit_intrinsic_bin_call(e, dest, intrinsic),
                 Intrinsic::IntMod => self.emit_intrinsic_bin_call(e, dest, intrinsic),
+                Intrinsic::IntNeg => self.emit_intrinsic_unary_call(e, dest, intrinsic),
+                Intrinsic::IntPlus => self.emit_expr(e.object.as_ref().unwrap(), dest),
```
dinfuehr: Could you please also invoke `emit_intrinsic_unary_call` here as well instead of handling the new intrinsics directly? Then we only have one implementation for them and the code should be more in line with the other intrinsics.

No need to change existing code - just the new intrinsics.

src/baseline/ast/expr.rs
soc: Agreed, will do. :-)

src/baseline/ast/expr.rs
soc: Ok, will do!

src/baseline/ast/expr.rs
soc: Also did the same for BoolNot.

soc:
Pushed!

dinfuehr:
Very cool, thanks!

PR #81, opened 2019-09-07
Turn modifier keywords into annotations

soc:
This splits the bare minimal set of keywords required to understand the
structure of the code from the larger set of keywords that act as
modifiers.

The core idea is to put as much of the language on equal footing to code
users could write, while also allowing the modifiers to benefit from
all the facilities annotations usually have, like the ability to have
parameters and defaults, as well as allowing migration and deprecation.

This is a limited implementation, as all supported annotations are
hard-coded and no syntax for defining annotations in Dora exists.

### DIFF
--- a/lib/dora-parser/src/lexer.rs
+++ b/lib/dora-parser/src/lexer.rs
@@ -363,7 +363,8 @@ impl Lexer {
                 } else {
                     TokenKind::Not
                 }
-            }
+            },
+            '@' => TokenKind::At,
 
             _ => {
                 return Err(MsgWithPos::new(
@@ -542,7 +543,7 @@ fn is_char_quote(ch: Option<char>) -> bool {
 }
 
 fn is_operator(ch: Option<char>) -> bool {
-    ch.map(|ch| "^+-*/%&|,=!~;:.()[]{}<>".contains(ch))
+    ch.map(|ch| "^+-*/%&|,=!~;:.()[]{}<>@".contains(ch))
         .unwrap_or(false)
 }
 
@@ -591,17 +592,9 @@ fn keywords_in_map() -> HashMap<&'static str, TokenKind> {
     keywords.insert("do", TokenKind::Do);
     keywords.insert("catch", TokenKind::Catch);
     keywords.insert("finally", TokenKind::Finally);
-    keywords.insert("abstract", TokenKind::Abstract);
-    keywords.insert("open", TokenKind::Open);
-    keywords.insert("override", TokenKind::Override);
     keywords.insert("defer", TokenKind::Defer);
-    keywords.insert("final", TokenKind::Final);
     keywords.insert("is", TokenKind::Is);
     keywords.insert("as", TokenKind::As);
-    keywords.insert("internal", TokenKind::Internal);
-    keywords.insert("optimize", TokenKind::Optimize);
-    keywords.insert("pub", TokenKind::Pub);
-    keywords.insert("static", TokenKind::Static);
     keywords.insert("spawn", TokenKind::Spawn);
     keywords.insert("const", TokenKind::Const);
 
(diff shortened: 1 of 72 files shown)


soc:
Mh, I thought I had an open issue with more rationale filed already ... let me know @dinfuehr if you want more reasoning that led to this PR.

soc:
```
$ rustup component add --toolchain nightly rustfmt
error: component 'rustfmt' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel 'nightly'

If you require these components, please install and use the latest successful build version,

which you can find at https://rust-lang-nursery.github.io/rust-toolstate.

After determining the correct date, install it with a command such as:

    rustup toolchain install nightly-2018-12-27
```

dinfuehr:
Looks good so far. Would subsequent PRs then remove the keywords? I'm okay merging this as is, but was wondering about your plans.

Me might want to standardise on one nightly-version at some point to prevent such errors.

soc:
> Would subsequent PRs then remove the keywords?

Do you mean in terms of allowing people to use the keywords as identifiers?

-> I think this is largely a matter of having some kind of namespacing/module system first, such that all annotations "live" somewhere. Then one would have to define which stuff is imported by default and polish a few smaller details. So I'd err on the conservative side for now, and keep them reserved until the module stuff is done.

... or in terms of making all annotations defined in code instead of having some hard-coded in the compiler?

-> I think this is largely an implementation/stylistic choice, in the end some of the core language annotations would need to be handed specially anyway.

dinfuehr:
I mean just getting rid of the keywords, so that code can use those keywords as identifiers in function/variable names, etc. A simple fix would just change `parse_annotations` to expect identifiers instead of the keywords.

It's your decision though, I personally wouldn't postpone this until both namespace/modules and annotation definitions are implemented.

soc:
I think this would be possible, but I think it would not necessarily free these keywords up as identifiers, because I wouldn't introduce a new namespace for annotations, which means annotation would live (probably) in the namespace for terms.

Let's try it!

dinfuehr:
True, but even in that case we need to get rid of the keywords eventually. I think that since we use `@`, we might be able to use separate namespaces for variables and namespaces.

soc:
>  we need to get rid of the keywords eventually

Agreed! I'm on it!

> I think that since we use @, we might be able to use separate namespaces for variables and namespaces.

I think it largely depends on how annotations are defined – if they were defined as

```
// special return type to signal annotations
fun deprecated(since: String): Annotation = ...
```
vs.
```
// annotation to define annotations
@annotation class deprecated(let since: String)
```
people might have different intuitions in which namespace they should be.

soc:
@dinfuehr Added commit that implements your advice of dropping the keywords.

lib/dora-parser/src/lexer.rs
```diff
@@ -582,6 +588,7 @@ fn keywords_in_map() -> HashMap<&'static str, TokenKind> {
     keywords.insert("static", TokenKind::Static);
     keywords.insert("spawn", TokenKind::Spawn);
     keywords.insert("const", TokenKind::Const);
+    keywords.insert("@", TokenKind::At);
```
dinfuehr: Sorry, didn't see this yesterday: Could you please parse `@` as an operator instead (in `read_operator`).

lib/dora-parser/src/parser.rs
```diff
@@ -472,18 +472,23 @@ impl<'a> Parser<'a> {
...
+                "abstract" => Modifier::Abstract,
+                "open" => Modifier::Open,
+                "override" => Modifier::Override,
+                "final" => Modifier::Final,
+                "internal" => Modifier::Internal,
+                "pub" => Modifier::Pub,
+                "static" => Modifier::Static,
+                "optimize" => Modifier::Optimize,
                 _ => {
                     break;
```
dinfuehr: I think we should throw an error here. We are already parsing `@<identifier>` here, so this should throw an error like `unknown annotation <identifier>`...

lib/dora-parser/src/lexer.rs
soc: Will do.

lib/dora-parser/src/parser.rs
soc: Will do!

lib/dora-parser/src/lexer.rs
soc: `read_operator` only returns the token itself, where should I parse the rest?

lib/dora-parser/src/parser.rs
soc: Done.

lib/dora-parser/src/lexer.rs
soc: Done.

dinfuehr:
Can you also please rebase your changes and run tests again?

soc:
@dinfuehr Done!

dinfuehr:
Thanks a lot! 👍

soc:
Cool, thanks!

PR #84, opened 2019-09-18
bytecode: add assert/throw support to bytecode

hzahradnik:
As discussed, the intrinsic assert is translated into the corresponding code, as it will be executed. This is similar to the approach Java is using.

### DIFF
--- a/src/bytecode/astgen.rs
+++ b/src/bytecode/astgen.rs
@@ -117,7 +117,7 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
             StmtVar(ref stmt) => self.visit_stmt_var(stmt),
             StmtWhile(ref stmt) => self.visit_stmt_while(stmt),
             StmtLoop(ref stmt) => self.visit_stmt_loop(stmt),
-            // StmtThrow(ref stmt) => {},
+            StmtThrow(ref stmt) => self.visit_stmt_throw(stmt),
             // StmtDefer(ref stmt) => {},
             // StmtDo(ref stmt) => {},
             // StmtSpawn(ref stmt) => {},
@@ -158,6 +158,11 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
         self.loops.pop();
         self.gen.emit_jump(start_lbl);
         self.gen.bind_label(end_lbl);
+    }
+
+    fn visit_stmt_throw(&mut self, stmt: &StmtThrowType) {
+        let exception_reg = self.visit_expr(&stmt.expr, DataDest::Alloc);
+        self.gen.emit_throw(exception_reg);
     }
 
     fn visit_stmt_if(&mut self, stmt: &StmtIfType) {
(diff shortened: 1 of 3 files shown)


src/bytecode/astgen.rs
```diff
@@ -284,9 +289,57 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
+        let fct_id = cls.constructor.expect("constructor is missing");
+        let fct = self.vm.fcts.idx(fct_id);
+        let fct = fct.read();
+
+        let arg_types = fct
+            .params_with_self()
+            .iter()
+            .map(|&arg| {
+                self.specialize_type_for_call(
+                    &CallType::CtorNew(cls_id, fct_id, TypeParams::Empty),
```
hzahradnik: I don't know, if it is correct to pass `TypeParams::Empty` to the function. What do you think?

dinfuehr: That should be fine, the error should not be generic.

src/bytecode/astgen.rs
```diff
@@ -284,9 +289,57 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
+                .into()
+            })
+            .collect::<Vec<BytecodeType>>();
+        let num_args = arg_types.len();
+
+        let start_reg = self.gen.add_register_chain(&arg_types);
+        let cls_id = specialize_class_id_params(self.vm, cls_id, &TypeParams::Empty);
+        self.gen.emit_new_object(start_reg, cls_id);
+        let error_string_reg = start_reg.offset(1);
+        let error_string_index = self.gen.add_string_const_pool("assert failed".to_string());
```
hzahradnik: I tried to come up with a more dynamic approach for the constructor arguments, but didn't find a solution. Mainly, because the arguments have to be set to specific values.

src/bytecode/astgen.rs
```diff
@@ -284,9 +289,57 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
         dest
     }
 
+    fn visit_expr_assert(&mut self, expr: &ExprCallType, _dest: DataDest) {
```
hzahradnik: The complete functions seems to be a little complicated or at least has some potential for optimization.

src/bytecode/astgen.rs
```diff
@@ -284,9 +289,57 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
+
     fn visit_expr_call(&mut self, expr: &ExprCallType, dest: DataDest) -> Register {
-        if let Some(_intrinsic) = self.get_intrinsic(expr.id) {
-            unimplemented!()
+        if let Some(intrinsic) = self.get_intrinsic(expr.id) {
+            match intrinsic {
+                Intrinsic::Assert => self.visit_expr_assert(expr, dest),
+                _ => unimplemented!(),
+            }
+            return Register::invalid();
```
hzahradnik: If I unterstand this correctly, assert is not a call with a return value, but (more or less) a statement with side effects. Therefore `Register::invalid()` should be correct.

hzahradnik:
For the record: As discussed in private, assert could also be written as a function in Dora prelude. This would replace the intrinsic call with an actual function call. This function could look like this:
```
fun assert(condition: Bool) {
    if(condition) {
        throw Error("assert failed");
    }
}
```
One drawback would be, that we have to deal differently with the stacktrace for the error.

I leave it for now as it is.

Issue #49, opened 2019-03-27
Plans for a minor release?


soc:
I don't really know what your goals for this project are, but what do you believe is required to release a minor version?

I'm fine if you don't care about this at all, my personal approach is "language quality and language popularity – pick one", so I'd be perfectly fine with keeping this language at ~2 users/contributors. :-)

But if you intend to do a release in the future, would you be interested in brainstorming and identifying what we think we need to ship a 0.0.x release? Maybe also considering none-core things like a website, documentation or IDE support (LSP)?

dinfuehr:
Not sure, haven't thought much about this since Dora for me is mostly about learning and probably also because there were just too many features missing.

Language Features I would like to see in a first release (although this might be more a wish list):
* better error messages (among other stuff report file names ;)
* implement HashMap (generics have to improved for this as well)
* some File I/O API
* implement structs+tuples
* ADT (enums) & pattern matching
* lambdas
* modules/packages with pub/public modifier
* operators like +=, -=, *=, ...
* allow if in expressions (you already proposed that issue)
* syntax highlighting for e.g. Visual Studio Code, VIM
* some documentation for the standard library

What do you think about releases and what features would you want to include?

soc:
@dinfuehr Thanks for your thoughts! Here is what I think:

> better error messages (among other stuff report file names ;)

Agree on this. I think making sure that all errors get a file name instead just those in the parser would be a reasonable sort-term goal.

> implement HashMap (generics have to improved for this as well)

I think we could survive without it for the next minor release, but on the other hand it has to be implemented anyway, so why not now.

> some File I/O API

I think it would make sense to restrict this to a very small, constrained scope, as I'm hoping to have some future API that is completely async (which would require something like `Future<T>`, which in turn requires improvements to generics).

> implement structs+tuples

Absolutely! I really want to have `Option` and `Result` types, and this would be a step toward that.

> ADT (enums)

Not sure this is critical, I really like that in most languages enums are their own sub-language with their own, special  hand-crafted syntax. I'd prefer first investigating how close we can get to use modules and classes/structs for this.

> pattern matching

I want this, but would avoid making this a goal, as pattern matching is very hard to implement if done right.
I have a design based on enhancing if-expressions that avoids introducing more syntax like (`match` or `switch` with their own special syntax rules). Some of the prerequisites for this aren't here yet, so I would postpone this for now.

> lambdas

[...]

PR #89, opened 2019-10-05
WIP: introduce github actions

riginding:
Hi, GitHub action is in beta, and everyone can join now.

I used github.com/denoland/deno GitHub action as an inspiration. It would be cool to provide tests and builds on (Linux, macOSX and windows), run linters and benchmarks, and create releases.

This basic GitHub action, just mirrors what the Travis build did.

I also wanted to add building and testing on macOSX, but I had some problems with capstone. Is the specific version (3.0.4) necessary?

What do you think about adding clippy? right now, dora's source code is not clippy conform.

Does it make sense to create releases with tags? It would be cool for users to download the binary [deno release page](https://github.com/denoland/deno/releases), or to see benchmark differences between versions (deno creates this [deno benchmark website](https://deno.land/benchmarks.html) on every commit).

Let me know if you think this is worthwhile and if I should continue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -148,7 +148,7 @@ dependencies = [
  "capstone 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "crossbeam-deque 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "docopt 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "dora-parser 0.1.0",
+ "dora-parser 0.0.1",
  "fixedbitset 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.62 (registry+https://github.com/rust-lang/crates.io-index)",
  "num_cpus 1.10.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


dinfuehr:
Hi!

Thanks for proposing this! @soc already brought up the idea of having [releases](https://github.com/dinfuehr/dora/issues/49). So I guess it makes sense to wait for feedback from him as well. Personally I am a bit worried about having releases too early (see https://news.ycombinator.com/item?id=21172980 what could happen). But I guess as long as we don't have a website and don't advertise the releases we should be fine. What do you think?

I don't think the one specific version is really required. README.md was already updated to use capstone from the OS repositories, travis was still using that specific version though. I've updated the travis-script now as well.

clippy sounds good and we should probably make use of it in the future. I haven't used it so far, but I guess it makes sense to fix issues reported by it and then enable it on in travis/github actions.

About the change: Do we really have to specify the rust version in the file? It would be awesome if we could update the rust-version just by changing the `rust-toolchain` file alone.

riginding:
Most of the negative comments could be avoided with proper communication. As far as I know Dora is a 'toy language' made for learning and trying things out, has this changed? Does Dora have a specific niche it caters to? Having releases would make it easier for everyone to just try Dora out, without compiling it yourself. (It could be that having the binaries is not really useful for anyone, and people are more excited to play with Dora's source code itself)

I will try setting up the GitHub actions without a specific capstone version.
Should I add `clippy` fixes in this PR, or should I create a new one?
We don't have to specify the rust version, that was a mistake.

dinfuehr:
True, I agree. Thinking about this, having releases would also make it easier to use Dora for scripting outside the project. Could we use that also to make "cargo install dora" possible?

I think it is better to not block this PR on fixing all clippy errors, so better to create PR(s) for clippy.

PR #93, opened 2019-10-09
Capstone not needed as external dependency

riginding:
I figured out that capstone is not needed as an external dependency anymore.
we are using capstone-rs, which is compiling capstone for us https://github.com/capstone-rust/capstone-rs/blob/master/capstone-sys/build.rs

The good news is we can also install with `cargo install dora`.
You only have to publish the project on crates.io https://doc.rust-lang.org/cargo/reference/publishing.html


dinfuehr:
Very cool, thanks!

PR #94, opened 2019-10-09
cannon: add support for various bytecodes

hzahradnik:
This PR adds support for various simple bytecode types, like arithmetic, compare and logical operations. Missing bytecodes include load, invoke, jump, new object and throw, as well some currently not existing bytecodes. I will implement them as soon as possible and add them to this PR or create a new one.


dinfuehr:
Cool, thanks!

Issue #62, opened 2019-04-07
Drop nil from the language


soc:
This proposal is based on the idea that we can remove and replace `nil`, which is untyped, can inhabit every reference type and therefore fail in unexpected places, with something safer.

###  Progress

- [x] Replace nil-producing array creation with alternatives.
- [x] Have an `Option` type.
- [ ] Runtime can represent options as references|nullpointer.
- [ ] Replace remaining nil-using code with alternatives.

---

The fundamental consideration is that it is much easier to disallow the entry points through which `nil`s can occur – than to allow them, and then try and make them safe and put the genie back into the bottle.

There are two ways to get a `nil` in Dora – either explicitly or implicitly. Let's dissect these cases separately:

### explicit nils

Explicit nils often occur as initial assignments to `var` in the source code and are later reassigned. Example:

```
var str: String = nil;
if condition {
  str = "foo";
} else {
  str = "bar";
}
```

Disallowing `nil` in such cases would require initializing them either to some "dummy" value (not that good) or to support if-expressions, such that the code above could be written as:

```
let str: String = if condition {
    "foo";
  } else {
    "bar";
  }
```

Note that this would enable us to also replace `var`s with `let`s in cases, where the binding was only mutable to allow initialization.

### implicit nils

Implict nils arise from creating arrays of reference types with a bare length, like in this example:

```
let array = Array[String](3);
```

Here my proposal is to further extend our factory functions (`arrayEmpty`, `arrayFill`) and make array construction via a bare length only accessible to these factory functions, such that every case where a user might want to create an unsafe, uninitialized array is catered by a safe function instead.

This means having at least (names are ad-hoc):

- `fun arrayEmpty[T]() -> Array[T]` as a replacement for code like `let array = Array[...](0)`
- `fun arrayFill<T>(len: Int, value: T) -> Array[T]` as a replacement for code like `let array = Array[...](2); array(0) = ...; array(1) = ...`
- `fun arrayGenerate[T](len: Int, func: Int -> T) -> Array[T]` as a replacement for code like `let array = Array[...](2); someInitFunction(array);`
- `fun arrayDefault[T : Default](len: Int) -> Array[T]` as a replacement for getting arrays that are initialized to the `T::default` for that type. (Probably `0`, `0.0` for numbers, `""` strings (?) and user-definable for their own types.)
[...]

PR #97, opened 2019-10-18
Drop nil from the parser, std lib and tests

soc:
Work toward resolving #62.


soc:
Current problem is replacing some instances of `nil` with an option type, but currently the `Option` type has onerous requirements (`Identity + Equals + Stringable` due to Generics limitations).

@dinfuehr Should I comment out `equals`/`has`/`contains`/`toString` to get rid of the bounds for now? Or do you see a better approach?

dinfuehr:
Nice! Oh right, didn't see that coming... No, unfortunately I don't see a better solution but I think it is fine to just move those functions out of the class for now. We already do a similar thing with some Array-functions (e.g. `arrayHas`) and it is very easy to clean this up in the future as soon as we support extension methods/impls.

dinfuehr:
Hmm, I think `defaultValue` is also able to produce `nil`s. Not sure what to do about it though, we can't really replace it in all use cases. Maybe we should limit its use to the stdlib as well and treat it as unsafe function.

dora/src/object.rs
```diff
@@ -517,7 +519,9 @@ impl Str {
 
             handle
         } else {
-            Ref::null()
+            let mut handle = str_alloc_heap(vm, 0);
+            handle.length = std::usize::MAX;
+            return handle;
```
soc: Will these hacks break anything in the short term until the code is replaced?

dinfuehr: Pretty sure the GC is going to crash on such an object. You could force a GC after this with `forceCollect` to test. 

What would you ideally do in such cases? I think throwing an exception makes sense here. This should already work from native [code](https://github.com/dinfuehr/dora/blob/master/dora/src/stdlib.rs#L105-L113).

dinfuehr: Might make sense to open separate PRs for these API changes.

soc: Basically I want to have an `Option` here (see the corresponding hack in https://github.com/dinfuehr/dora/pull/97/files#diff-57504af7150824bb788bd6c48e08ca9e) but I'm unsure whether it is a good idea to assume a specific memory representation.

If getting a valid Dora Option from Rust code is too inconvenient at the moment, maybe it would be possible to make the GC aware of this hack to keep things going.

dinfuehr: We just can't add such a hack to the codebase. Especially since it doesn't block this work at all. There are perfectly fine alternatives: 1) throw an exception 2) return empty string 3) add support for Option for native code 4) rewrite function in Dora.

soc: You are right. We could return an empty string if it failed, and check whether the input was also empty: If it was, everything is fine, otherwise we encountered an error. Does that sound good to you?

soc: @dinfuehr I removed the ugly hack in https://github.com/dinfuehr/dora/pull/97/commits/ef2c530aa078d6ad4d720c887d741dffacad85be. Does it look good do you?

soc:
@dinfuehr I think I'd introduce a private function that allocates an array of a certain size without initialization and without requiring any constraints. This would deal with the vector case and related ones. What do you think?

dora/src/object.rs
dinfuehr: I would've preferred the exception, since it makes the failure explicit but this is also okay.

dinfuehr:
Isn't that just what the array-constructor does?

soc:
You are right – I was under the wrong impression that the array constructor also gained some constraints already.

dora/src/object.rs
dinfuehr: Yes, that's possible. That's why I linked the [throw_native](https://github.com/dinfuehr/dora/blob/master/dora/src/stdlib.rs#L116-L126) function above. This function doesn't do anything useful, it's only here for testing this functionality. But if you pass `true` as argument it is going to throw an exception from native code.

dinfuehr: (Sorry for not explicitly mentioning that)

soc: Ah, now I saw it! Thanks, I will revise the code accordingly!

soc:
Last failing tests:
- `tests/native-throw2.dora.main... failed (expected success (0 expected but test returned 104))`:
  The problem is that the native code creates an exception with a `nil` String – I'd replace it with an empty string instead, @dinfuehr?
- `tests/bench/splay1.dora.main... failed (expected success (0 expected but test returned 104))` and `tests/bench/splunc1.dora.main... failed (expected success (0 expected but test returned 104))`: Same problem, probably some mistake while migrating the code, will need to look into this.

dinfuehr:
Yes, just replace it with an empty string or some string like "test exception". If you need help, just reach out 👍

soc:
> Yes, just replace it with an empty string or some string like "test exception".

Alright, will do!

> If you need help, just reach out +1

Thanks! If you completely ran out of work, you could have a look at the other test failure (the splunc/splay) one. It's probably just some silly mistake while migrating it, but a fresh pair of eyes could help.

dinfuehr:
Ok, will try to take a look in the next few days!

dora/src/stdlib.rs
```diff
@@ -117,7 +117,7 @@ pub extern "C" fn throw_native(val: bool) {
     if val {
         handle_scope(|| {
             let vm = get_vm();
-            let obj = alloc_exception(vm, Ref::null());
+            let obj = alloc_exception(vm, Str::empty(vm));
```
soc: I thought this would resolve the issue, but it didn't. Did I miss anything here?

dinfuehr:
Not sure what to do here, I guess `nil` is harder to remove than I anticipated 😞 ...  Another problem I've encountered are global variables: Right now those cannot have initializers so they are always `nil` in case of references. I would like to solve this similar to Swift where global variables are initialized lazily.

I guess this PR can't get rid of `nil` entirely, but we should try to get your changes in since they remove a lot of uses of nil. My proposal: rebase this PR and remove nil where it is simple enough. Do not remove the `nil` keyword for now. What do you think about this?

BTW: Not sure if you have seen that I've updated `Option` to a class hierarchy. This was my testcase for making sure generics work with inheritance.

soc:
Hey @dinfuehr!

Regarding

> updated Option

is the change to underscore-case for some methods in Option intentional (`is_some` instead of `isSome`, `is_none` instead of `isNone`?)?

dinfuehr:
> is the change to underscore-case for some methods in Option intentional (`is_some` instead of `isSome`, `is_none` instead of `isNone`)?

I was never really consistent with that when writing Dora code. That's something I actually wanted to ask you about, I guess we can pick either camel or snake case and then just stick with it.

soc:
> I was never really consistent with that when writing Dora code. That's something I actually wanted to ask you about, I guess we can pick either camel or snake case and then just stick with it.

The impression I got from Rust is that under_score_case makes people subconsciously avoid method names with more than 1-2 words (even where it ends up with a substantially less descriptive name), because the `_` is adding quite some visual overhead.

I'd prefer camelCase; I'll create a PR if your ok with this.

soc:
@dinfuehr PR would be https://github.com/dinfuehr/dora/pull/120.

soc:
#128.

Issue #99, opened 2019-10-25
Drop break and continue control-flow keywords


soc:
Extracted from https://github.com/dinfuehr/dora/issues/39:

I'd like to get rid of those two control-flow keywords. Reasoning:

- The alternative of using an additional method feels only a slightly bit clunkier to write and vastly easier to read and understand when coming back after a month.¹
- In my experience the implementation complexity and the mental complexity for users has never been worth the "convenience".
- There aren't much reasons for `break` and `continue` to exist in general – it feels like this is something that got copied from C and keeps getting copied without anyone questioning it much. I certainly think that if `break` and `continue` didn't exist today, we wouldn't invent it.

Prior art:

One language that did away with `continue` is Scala; and in the last 10 years I never heard a single complaint about it. (`break` in Scala is done as a library (that throws exceptions) – I rather not go this way either.)

---

¹ I ported Java's `java.time` implementation from Java to Scala and had to deal with a lot of `break`s and `continue`s, and the replacements were never too bad. Replacing `break`s is almost trivial – `continue` is a bit more involved, but barely ever used.

The biggest trouble was always understanding what the `break` or `continue` was trying to do in the first place – a problem we won't have without `break` and `continue` in the first place.

soc:
One additional reason is that it prevents building pure library-defined resource handling:

Consider

```
fun use[R : Closeable, T](resource: R, function: T -> Unit) {
    function();
    resource.close();
}
while ... {
  use(someFile, |x| { printFile(someFile); if someFile.isEmpty() { break; } } );
}
```

"Use `do ... finally ...` in the `use` function" would be the obvious answer, but in the long term I'd like to adopt a Rust-like approach where exceptions are only used for fatal errors that require aborting the program.

Not having `break`, `continue` and exceptions for control flow would allow us to replace `finally` by a pure library implementation.

dinfuehr:
Haven't looked at the PR so far.

Don't get your argument here though and would be curious what I am missing here: break/continue/return only work inside the current function or closure. So the example above would fail to compile (is there missing something?). How would you implement resource handling "as a library" in Dora?

soc:
Feel free to substitute `break` with `throw`; but I agree with you that it is the least compelling point listed here. Feel free to disregard it.

dinfuehr:
I don't get why `break`/`continue` are different in this regard. It seems that even `return` would break resource handling already. How would you otherwise ensure that the `close`-function is run (without finally or defer)?

[...]

PR #100, opened 2019-10-26
Drop break/continue; drop loop

soc:
I gave #99 a try, what do you think?


dinfuehr:
Sorry for taking so long to look into this PR... Not sure about this change though. I don't really see the advantages so far, without break/continue you'd be forced to add this additional mutable variable and possibly more if-conditionals. The advantages you see is purely stylistic, not sure I agree since break/continue is very well understood by many programmers and it is perfectly possible to write (un)maintable code with or without them. 

I am also a bit worried, since I now want to write some Dora code in dora-boots. I might miss those keywords there. What also makes me wonder is that Scala for example at least has break, we would have neither. Right now in Dora we don't yet have some features that would make it easier to come by without them, like filter/map/... functions. So I would propose to revisit this PR again later. After all it is much easier to remove features than adding them. What do you think?

soc:
Closing; drop break/continue will be revisited at a later date, drop loop is being done in https://github.com/dinfuehr/dora/pull/186.

PR #101, opened 2019-10-31
chore: include stdlib

riginding:
Dora now has the stdlib baked in.

I leaned heavily on this SO Answer: https://stackoverflow.com/a/50554062/1354814 to include all files in stdlib.

I had to move the stdlib into the Dora package because it was not possible to access it outside of the Dora package.

It was hard to verify that my changes are working because the test suite is using `dora/src/test.rs` which has a hardcoded path of the stdlib files. Please advise how I can test if it actually works :smile:


dora/build.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    writeln!(&mut stdlib, r#"["#,)?;
+
+    for f in fs::read_dir(SOURCE_DIR)? {
+        let f = f?;
+
+        if !f.file_type()?.is_file() {
+            continue;
+        }
+
```
riginding: I think I should add a guard, that the to-be-included-files end with `.dora`

dora/src/driver/start.rs
```diff
@@ -103,31 +103,39 @@ pub fn start(content: Option<&str>) -> i32 {
...
-        .unwrap_or_else(|| "stdlib".to_string());
+        .clone();
 
-    parse_dir(&stdlib, vm, ast).and_then(|_| {
-        if fuzzing {
-            return parse_str(content.unwrap(), vm, ast);
+    if let Some(stdlib) = stdlib_dir {
+        parse_dir(&stdlib, vm, ast)?;
+    } else {
+        for (_, data) in STDLIB {
```
riginding: we have a tuple of ( filename, content ),
we are not using the filename right now, but i think in the future this is necessary for better error reporting.

dora/build.rs
dinfuehr: Yes, I agree. Out of interest: Does this also include files recursively?

dora/src/driver/start.rs
```diff
@@ -103,31 +103,36 @@ pub fn start(content: Option<&str>) -> i32 {
     code
 }
 
+const STDLIB: &[(&str, &str)] = &include!(concat!(env!("OUT_DIR"), "/stdlib.rs"));
```
dinfuehr: Could you please rename this file to something else, e.g `dora_stdlib_bundle.rs`? We already have a file `stdlib.rs`, this might otherwise be confusing...

dora/build.rs
dinfuehr: Yes, please add this flag.

dora/src/driver/start.rs
dinfuehr: Yes we should do that. You don't have to do this in this PR though, I can do that later as well if you want.

dora/build.rs
```diff
@@ -0,0 +1,55 @@
...
+        copy(root_path.join(f.path()), out_path.join(f.path()))?;
+
+        writeln!(
+            &mut stdlib,
+            r#"("{name}", include_str!("{name}")),"#,
```
dinfuehr: Very cool, didn't think of generating Rust-code for this 👍

dora/src/driver/start.rs
dinfuehr: BTW If you want to tackle this, shouldn't be too hard: Reader::from_string should take another argument (the filename).

dora/build.rs
riginding: > Yes, I agree. Out of interest: Does this also include files recursively?

No, it does not. I can add it in a different PR

riginding: I added a check now

dora/src/driver/start.rs
riginding: I created a new `Reader::from_bundle` function, do you prefer changing the `Reader::from_string` function?

dinfuehr: I would actually prefer changing `Reader::from_string` but I would also be fine with your current approach.

riginding: I think there is no function overloading and also no default arguments in Rust, so I am not sure how the signature should look like:

`Reader::from_string(filename: &str, content: &str)`
`Reader::from_string(filename: Option<&str>, content: &str)`

dinfuehr: Just always expect the filename.

dinfuehr: so `Reader::from_string(filename: &str, content: &str)`

dinfuehr:
Thanks!

PR #102, opened 2019-11-01
cannon: support for jump

hzahradnik:
This PR adds support for the bytecodes for Jump, JumpIfTrue & JumpIfFalse. To resolve forward jumps, I added a Map from bytecode index to address, which is then used to create a label at a specific position. 

Additionally I fixed two small errors:
- a if expession, which returns a Unit, crashed the bytecode generator
- a return inside the code did not exit the function

### DIFF
--- a/dora/src/baseline/cannon/codegen.rs
+++ b/dora/src/baseline/cannon/codegen.rs
@@ -19,7 +19,7 @@ use crate::bytecode::opcode::Bytecode;
 
 pub struct ForwardJump {
     label: Label,
-    bytecode_idx: usize,
+    bytecode_idx: BytecodeIdx,
 }
 
 pub struct CannonCodeGen<'a, 'ast: 'a> {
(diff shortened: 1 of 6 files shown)


dora/src/baseline/cannon/codegen.rs
```diff
@@ -745,11 +830,13 @@ impl<'a, 'ast> CodeGen<'ast> for CannonCodeGen<'a, 'ast> {
                 | Bytecode::RetFloat(src)
                 | Bytecode::RetDouble(src) => self.emit_return_generic(&bytecode, *src),
                 Bytecode::RetVoid => {}
-                _ => panic!("bytecode {:?} not implemented", btcode),
+                _ => self.emit_epilog(),
```
dinfuehr: Why this? Why not crash for unknown bytecodes?

dora/src/baseline/cannon/codegen.rs
```diff
@@ -587,6 +597,55 @@ where
...
+            }
+        }
+    }
+
+    fn emit_jump(&mut self, _bytecode: &BytecodeFunction, bytecode_idx: &BytecodeIdx) {
+        let lbl = self.asm.create_label();
+        self.asm.jump(lbl);
+
+        let BytecodeIdx(bytecode_idx) = bytecode_idx;
+        match self.bytecode_to_address.get(&bytecode_idx) {
```
dinfuehr: Right now, `bytecode_idx` is always increasing by 1. In the future this will be the offset in the bytecode stream. That's why I think it would be safer to shield against bugs with wrong offsets: `if target < current_position { self.bytecode_to_address.get(&offset).expect("jump with wrong offset") } else { /* forward jump*/ }`

dora/src/baseline/cannon/codegen.rs
```diff
@@ -49,6 +56,9 @@ pub struct CannonCodeGen<'a, 'ast: 'a> {
 
     pub cls_type_params: &'a TypeList,
     pub fct_type_params: &'a TypeList,
+
+    pub bytecode_to_address: HashMap<usize, usize>,
```
dinfuehr: I think it would be safer if the HashMap would be `HashMap<BytecodeIdx, usize>`.

dora/src/baseline/cannon/codegen.rs
```diff
@@ -587,6 +597,55 @@ where
             .store_mem(bytecode_type.mode(), Mem::Local(offset), REG_RESULT.into());
     }
 
+    fn emit_jump_if(
+        &mut self,
+        bytecode: &BytecodeFunction,
+        src: Register,
+        bytecode_idx: &BytecodeIdx,
```
dinfuehr: I think you don't need to pass this as reference.

dora/src/baseline/cannon/codegen.rs
```diff
@@ -587,6 +597,55 @@ where
             .store_mem(bytecode_type.mode(), Mem::Local(offset), REG_RESULT.into());
     }
 
+    fn emit_jump_if(
+        &mut self,
+        bytecode: &BytecodeFunction,
```
dinfuehr: Not in this PR, but move the `BytecodeFunction` into the struct. No need to pass this reference to so many functions.

dora/src/baseline/cannon/codegen.rs
```diff
@@ -587,6 +597,55 @@ where
...
+        src: Register,
+        bytecode_idx: &BytecodeIdx,
+        op: CondCode,
```
dinfuehr: `CondCode` allows all kinds of comparison operators, I guess it is better to just pass in `true` and `false` instead.

dora/src/baseline/cannon/codegen.rs
dinfuehr: Would be awesome if these two functions could share this code as well.

dora/src/baseline/cannon/codegen.rs
hzahradnik: This was a mistake.

dora/src/baseline/cannon/codegen.rs
hzahradnik: For now, I only moved the code to a separate function, but didn't add the check. I will revisit this tomorrow.

PR #105, opened 2019-11-06
parser: added function modifier for cannon

hzahradnik:
As discussed in Hangouts, I have added a modifier to mark functions explicitly to be executed with cannon. This should allow us to test cannon better, even if it doesn't support all of the functionality.


dinfuehr:
cool!

PR #107, opened 2019-11-06
Add modules to AST and parser

soc:
These are the first round of changes salvaged from #88.

(My approach now would be to add these things step-by-step as you proposed, such that the PR doesn't become completely full of conflicts.)


dinfuehr:
Thanks!

PR #109, opened 2019-11-08
Add modules to VM

soc:
This is the last part of the changes from #88.

After this, I will work on the approach you outlined, beginning with the first point:

> The first PR could simply introduce modules and only allow static functions in it. Making <module>::<fctname>() calls work should be relatively easy. Do not allow modules and classes with the same name for now.

### DIFF
--- a/dora/src/module.rs
+++ b/dora/src/module.rs
@@ -1,0 +1,108 @@
+use parking_lot::RwLock;
+use std::sync::Arc;
+
+use crate::class::ClassId;
+use crate::field::{Field, FieldDef};
+use crate::size::InstanceSize;
+use crate::ty::BuiltinType;
+use crate::utils::GrowableVec;
+use crate::vm::{FctId, FileId, TraitId, VM};
+use crate::vtable::VTableBox;
+
+use dora_parser::interner::Name;
+use dora_parser::lexer::position::Position;
+
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
+pub struct ModuleId(usize);
+
+impl ModuleId {
+    pub fn max() -> ModuleId {
+        ModuleId(usize::max_value())
+    }
+}
+
+impl From<ModuleId> for usize {
+    fn from(data: ModuleId) -> usize {
+        data.0
+    }
+}
+
+impl From<usize> for ModuleId {
+    fn from(data: usize) -> ModuleId {
+        ModuleId(data)
+    }
+}
+
+impl GrowableVec<RwLock<Module>> {
+    pub fn idx(&self, index: ModuleId) -> Arc<RwLock<Module>> {
+        self.idx_usize(index.0)
+    }
+}
+
+pub static DISPLAY_SIZE: usize = 6;
+
+#[derive(Debug)]
+pub struct Module {
+    pub id: ModuleId,
+    pub file: FileId,
+    pub pos: Position,
+    pub name: Name,
+    pub ty: BuiltinType,
+    pub parent_class: Option<ClassId>,
+    pub internal: bool,
+    pub internal_resolved: bool,
+    pub has_constructor: bool,
+
+    pub constructor: Option<FctId>,
+    pub fields: Vec<Field>,
+    pub methods: Vec<FctId>,
+    pub virtual_fcts: Vec<FctId>,
+
+    pub traits: Vec<TraitId>,
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct ModuleDefId(usize);
+
+impl ModuleDefId {
+    pub fn to_usize(self) -> usize {
+        self.0
+    }
+}
+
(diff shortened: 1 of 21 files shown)


dora/src/module.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub ty: BuiltinType,
+    pub parent_class: Option<ClassId>,
+    pub internal: bool,
+    pub internal_resolved: bool,
+    pub has_constructor: bool,
+
+    pub constructor: Option<FctId>,
+    pub fields: Vec<Field>,
+    pub methods: Vec<FctId>,
+    pub virtual_fcts: Vec<FctId>,
```
dinfuehr: Do we really need `virtual_fcts`, `traits` and `impls` in a module? Maybe we've discussed that before already though.

soc: I will check, I would have removed the unnecessary things after the next phase anyway, I just wasn't sure what I would actually need before implementing it.

Looking at it right now:
- trait: Yes, modules should be able to implement traits.
- impl: Unneeded, probably.
- virtual_fcts: No idea, does the Vec include virtual funs inherited from the parent class?

soc: Removed `impl`!

soc:
With the second commit this gets us right up to the baseline and fails in `InfoGenerator`.

Do you think the two commits (either individually or combined) are self-contained enough to be merged to avoid merge conflicts later?

dora/src/semck/moduledefck.rs
```diff
@@ -0,0 +1,363 @@
...
+    #[test]
+    fn let_field_without_initialization() {
+        err(
+            "module Foo { let a: Int; }",
+            pos(1, 14),
+            SemError::LetMissingInitialization,
+        );
+    }
+
+    //#[test] // should fail but doesn't
```
dinfuehr: I think it would be better to simply `#[ignore]` the tests that don't work yet.

dinfuehr:
Looks good. To be honest I am still not sold on this feature but we can definitely merge this as-is. Although it needs to be rebased since some stuff was moved around - sorry for that again. Next time feel free to ping me after a few days already...

I am a bit sceptical about the singleton use of the `module` keyword though to be honest. I am worried that the namespace and singleton features of the `module` keyword will sooner or later conflict in some way. Not sure the singleton feature needs to be in the language at all, especially since in the long-run I would like to get rid of inheritance.

soc:
@dinfuehr I agree with you on the concern about inheritance, it's a feature that adds a lot of complexity to many things.

Is your move from `Option` as a single class to a class hierarchy orthogonal to this? Or do you have a restricted form of inheritance in mind that would keep such a use case working?

dinfuehr:
I thought about implementing enums/ADTs as a class hierarchy initially first. Then we would need inheritance+generics for classes. There are some issues with enums stored as structs with respect to garbage collection. Other motivations were getting rid of defaultValue and simply curiosity.

soc:
I'm also in favor of using the existing machinery for enums.

My vision is to cut the enum functionality down to its core such that it isn't another language feature with its own keyword, but just another annotation that can be added to sealed hierarchies, and doesn't do much more than instruct the VM to generate a few convenience methods (e. g. to get all values, lookup up a enum member by string, etc.).

My experience from other languages like Scala and Rust is that having enums as a separate feature often incurs quirks and mismatches between the "enum feature" and existing language facilities that enable building "enum-like" things.

Using existing language features is a bit more verbose at the declaration-site, but does not have any costs at the use-site, which I think is more important to look at.

In the simplest case, I would imagine an `@enum` annotation be used like this:

```
@enum
module Pets {
  module Cat
  module Dog
}
```

... and more elaborate cases like this:

```
@enum
module Pets {
  module Cat : Pet("meow")
  module Dog : Pet("woof")
}
class Pet(let sound: String)
```

Usage:

```
Pets.valueOf("Cat") // returns module Cat
Pats.values()     // returns Array(Cat, Dog)
```

I wouldn't even allow `class` values for enums, such that an enum can be a fixed enumeration of known values.

Scala tried to having it both ways and it's a bit of a mess: https://scastie.scala-lang.org/1cNHCOijQ22ESGXYRPYLjA

dinfuehr:
> E. g. an enum `E` has members `X` and `Y`, and a function only makes sense for `X` – without enum members as types there is no way to restrict the function's input to only `X`, one is forced to accept all of `E`'s members.

Sounds interesting. In Rust I would solve this by introducing a struct for X. Would every member be a type or is there a way to opt-in to this? How would `Option` be implement in this scheme?

> * ADTs as class hierarchies.

I was thinking about implementing them using class hierarchies, but for the developer this was supposed to be an implementation detail. Right now I think I've understood how to implement them as value types, which would be quite important for performance.

soc:
@dinfuehr Rebased. Sorry for the delay, changes were pretty big.

dinfuehr:
Nice, thanks!

PR #112, opened 2019-11-10
cannon: float operations

hzahradnik:
Based on PR #111 

Adds arithmetic (-,+,/,*) and compare operations (==,!=,<,<=,>=,>) for Float to the bytecode generator and cannon. Includes a simple test for these operations.

Some small fixes in the dump() function of the bytecode generator are also included. The strings didn't always represent the operation correctly. ;)


hzahradnik:
I didn't activate them in this PR, because I thought it would be better to create a separate PR entirely focused on tests. If you want, I will include it in this PR, otherwise I will create the new PR tomorrow.

dinfuehr:
No, that's not necessary. Just wanted to ask.

PR #114, opened 2019-11-23
cannon: small fix in load global fields (+enable tests)

hzahradnik:
This PR adds some tests for cannon, using the new function modifier. As a result of this tests, I found a bug when loading global float fields.


hzahradnik:
I changed the commented line and rebased the PR.

dinfuehr:
Cool, thanks!

PR #116, opened 2019-12-22
Fix #115: Implement Sortable for Float and Double

### LINKED ISSUE #115: Implement Sortable for Float/Double

soc:
Supposed to fix #115.

### DIFF
--- a/dora/src/masm/x64.rs
+++ b/dora/src/masm/x64.rs
@@ -269,6 +269,47 @@ impl MacroAssembler {
         }
     }
 
+    pub fn float_cmp_total(
+        &mut self,
+        src_mode: MachineMode,
+        dest_mode: MachineMode,
+        dest: Reg,
+        lhs: FReg,
+        rhs: FReg,
+    ) {
+        let bits_to_flip = match dest_mode {
+            MachineMode::Int32 => 31,
+            MachineMode::Int64 => 63,
+            _ => unimplemented!(),
+        };
+
+        self.float_as_int(dest_mode, RAX, src_mode, lhs);
+        self.float_as_int(dest_mode, RCX, src_mode, rhs);
+        self.float_as_int(dest_mode, RDX, src_mode, lhs);
+        self.float_as_int(dest_mode, RSI, src_mode, rhs);
+
+        self.load_int_const(dest_mode, REG_TMP1, bits_to_flip);
+        self.int_sar(dest_mode, RDX, RDX, REG_TMP1);
+        self.load_int_const(dest_mode, REG_TMP1, bits_to_flip);
+        self.int_sar(dest_mode, RSI, RSI, REG_TMP1);
+        self.load_int_const(dest_mode, REG_TMP2, 1);
+        self.int_shr(dest_mode, RDX, RDX, REG_TMP2);
+        self.int_shr(dest_mode, RSI, RSI, REG_TMP2);
+        self.int_xor(dest_mode, RDX, RDX, RAX);
+        self.int_xor(dest_mode, RSI, RSI, RCX);
+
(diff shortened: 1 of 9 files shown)


dora/src/masm/x64.rs
```diff
@@ -269,6 +269,41 @@ impl MacroAssembler {
...
+        self.int_xor(dest_mode, RCX, RCX, RCX);
+        self.cmp_reg(dest_mode, RDX, RSI);
+        self.set(RCX, CondCode::NotEqual);
+        self.load_int_const(dest_mode, RAX, -1);
+        let x64 = match dest_mode {
+            MachineMode::Int32 => 0,
+            MachineMode::Int64 => 1,
+            _ => unimplemented!(),
+        };
+        asm::cmov(self, x64, dest, RCX, CondCode::GreaterEq);
```
soc: This implementation is derived from https://godbolt.org/z/xgDe2q, which in turn is based on [__totalorderf](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/ieee754/flt-32/s_totalorderf.c;h=71ea5968ab3eadf575c42861f83a17092f803723;hb=0a8262a1b269f8f1602933248e69dfb041aab7a2):
```
        movd    eax, xmm0
        movd    ecx, xmm1
        mov     edx, eax
        sar     edx, 31
        mov     esi, ecx
        sar     esi, 31
        shr     edx
        xor     edx, eax
        shr     esi
        xor     esi, ecx
        xor     ecx, ecx
        cmp     edx, esi
        setne   cl
        mov     eax, -1
        cmovge  eax, ecx
        ret
```

It looks like I made some mistake translating it though, see the test file with comments.

tests/srt.dora
```diff
@@ -0,0 +1,19 @@
...
+    assert((0.0F/0.0F).sortsAs(0.0F/0.0F) == -1); //wrong
+
+    assert(0.0.sortsAs(0.0) == -1);    //wrong
+    assert(-0.0.sortsAs(-0.0) == 1);   //wrong
+    assert(1.0.sortsAs(1.0) == 1);     //wrong
+    assert(0.0.sortsAs(1.0) == -1);
+    assert(1.0.sortsAs(0.0) == 1);
+    assert(-0.0.sortsAs(0.0) == 1);    //wrong
+    assert(0.0.sortsAs(-0.0) == -1);   //wrong
+    assert((0.0/0.0).sortsAs(0.0/0.0) == -1); //wrong
```
soc: Can't really see a rhyme or reason in which comparison returns the right or wrong result ...

dora/src/masm/x64.rs
```diff
@@ -1082,6 +1123,13 @@ pub struct ForwardJump {
 mod tests {
     use super::*;
 
+    #[test]
+    fn test_float_cmp_total() {
```
dinfuehr: I don't think it is necessary to test the assembly-output here. IMHO your tests in tests/srt.dora are good enough. Otherwise we would have here a long sequence of machine code bytes, which is incredibly hard to update/read/maintain.

dinfuehr:
Not sure what the problem is here (although I haven't looked into the generated assembly code at all). What usually helps here: emit the `debug`-instruction right before the assembly-code of the intrinsic, then run your test case in gdb. Execution will then stop right before the intrinsic and you can run each instruction manually and check its result.

OTOH if this is too complicated, you could also write this function in Dora for now. I know this function might get hot, but we can always replace the Dora version with a more optimized version later. I want to fix ARM64 support sometime in the next few months as well and the Dora version would make this easier :D

soc:
@dinfuehr Superseded by https://github.com/dinfuehr/dora/pull/119 (at least for now – I still want to intrinisfy it, but for now let's use an implementation in Dora to clear up the PR queue).

PR #121, opened 2020-01-01
Add {Int,Long,Float,Double,Bool}#toStringBinary

soc:
This change adds a helper function that's helpful when working with bits.

This is certainly not the most elegant, elaborate or flexible approach
(compared to languages that support arbitrary bases or compared to a
typeclass-based design that enables a maximum of abstraction), but it is
simple and straight-forward and should last until a better solution is
required.

### DIFF
--- a/dora/stdlib/Double.dora
+++ b/dora/stdlib/Double.dora
@@ -4,6 +4,8 @@
   @internal fun toFloat() -> Float;
   @internal fun toString() -> String;
 
+  fun toStringBinary() -> String = String::toStringBinary(self.asLong(), Double::bits());
+
   @internal fun asLong() -> Long;
 
   @internal fun equals(rhs: Double) -> Bool;
@@ -35,4 +37,7 @@
   @internal fun sqrt() -> Double;
 
   fun hash() -> Int = self.asLong().toInt();
+
+  @static fun bits() -> Int = 64;
+  @static fun bytes() -> Int = 8;
 }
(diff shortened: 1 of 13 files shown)


dora/stdlib/Double.dora
```diff
@@ -4,6 +4,16 @@
   @internal fun toFloat() -> Float;
   @internal fun toString() -> String;
 
+  fun toStringBinary() -> String {
```
dinfuehr: Couldn't we share all those implementations in a single function (apart from bool)? Might also be a good use-case for the StringBuffer class.

dinfuehr:
Agreed, hex-strings would be nice as well.

soc:
> Agreed, hex-strings would be nice as well.

Yes, had that in mind too!

dora/stdlib/Double.dora
soc: Had considered it, but I didn't want to have a "public" method somewhere that was only supposed to be used internally. Where would you place it?

dinfuehr: I suppose in any of these files is good enough for now. That this function is effectively `public` for now is definitely unfortunate, let's hope we will have some form of encapsulation sooner or later.

soc: Alright, I updated the PR.

dinfuehr:
Cool, thanks!

PR #127, opened 2020-01-04
cannon: support creation of objects

hzahradnik:
This PR allows to call the constructor of a class to create a new object. As a consequence StoreField* are now supported.

### DIFF
--- a/dora/src/bytecode/generator.rs
+++ b/dora/src/bytecode/generator.rs
@@ -230,7 +230,7 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
             ExprLitBool(ref lit) => self.visit_expr_lit_bool(lit, dest),
             ExprIdent(ref ident) => self.visit_expr_ident(ident, dest),
             ExprCall(ref call) => self.visit_expr_call(call, dest),
-            // ExprDelegation(ref call) => {},
+            ExprDelegation(ref call) => self.visit_expr_delegation(call, dest),
             ExprSelf(ref selfie) => self.visit_expr_self(selfie, dest),
             // ExprSuper(ref expr) => {},
             ExprNil(ref nil) => self.visit_expr_nil(nil, dest),
@@ -486,6 +486,77 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
         } else {
             return_reg
         }
+    }
+
+    fn visit_expr_delegation(&mut self, expr: &ExprDelegationType, dest: DataDest) -> Register {
+        let call_type = self.src.map_calls.get(expr.id).unwrap().clone();
+        let fct_id = call_type.fct_id().unwrap();
+
+        let fct = self.vm.fcts.idx(fct_id);
+        let fct = fct.read();
+
+        let callee_id = if fct.kind.is_definition() {
+            unimplemented!()
+        } else {
+            fct_id
+        };
+
+        let callee = self.vm.fcts.idx(callee_id);
+        let callee = callee.read();
+
+        if let FctKind::Builtin(_intrinsic) = callee.kind {
+            unimplemented!()
+        }
+
+        let return_type = if dest.is_effect() {
+            BuiltinType::Unit
+        } else {
+            self.specialize_type_for_call(&call_type, callee.return_type)
+        };
+        let arg_types = callee
+            .params_with_self()
+            .iter()
+            .map(|&arg| self.specialize_type_for_call(&call_type, arg).into())
+            .collect::<Vec<BytecodeType>>();
+        let num_args = arg_types.len();
+
+        let return_reg = if return_type.is_unit() {
+            Register::invalid()
+        } else {
+            self.ensure_register(dest, return_type.into())
+        };
+
+        let start_reg = if num_args > 0 {
+            self.gen.add_register_chain(&arg_types)
+        } else {
+            Register::zero()
+        };
+
+        let arg_start_reg = if callee.has_self() {
+            let self_id = self.src.var_self().id;
+            let self_reg = self.var_reg(self_id);
+
+            self.gen.emit_mov_ptr(start_reg, self_reg);
+            start_reg.offset(1)
+        } else {
+            start_reg
+        };
+
+        for (idx, arg) in expr.args.iter().enumerate() {
+            let arg_reg = arg_start_reg.offset(idx);
+            self.visit_expr(arg, DataDest::Reg(arg_reg));
+        }
+
+        match *call_type {
+            CallType::Ctor(_, _) => {
+                self.gen
+                    .emit_invoke_direct_void(callee_id, start_reg, num_args);
+            }
+
+            _ => unreachable!(),
+        }
+
+        return_reg
     }
 
     fn visit_expr_nil(&mut self, _nil: &ExprNilType, dest: DataDest) -> Register {
--- a/dora/src/bytecode/generator.rs
+++ b/dora/src/bytecode/generator.rs
@@ -489,6 +489,7 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
     }
 
     fn visit_expr_delegation(&mut self, expr: &ExprDelegationType, dest: DataDest) -> Register {
+        assert!(dest.is_effect());
         let call_type = self.src.map_calls.get(expr.id).unwrap().clone();
         let fct_id = call_type.fct_id().unwrap();
 
@@ -496,11 +497,7 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
         let callee = self.vm.fcts.idx(callee_id);
         let callee = callee.read();
 
-        let return_type = if dest.is_effect() {
-            BuiltinType::Unit
-        } else {
-            self.specialize_type_for_call(&call_type, callee.return_type)
-        };
+        assert!(callee.return_type.is_unit());
         let arg_types = callee
             .params_with_self()
             .iter()
@@ -522,7 +519,6 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
             self.visit_expr(arg, DataDest::Reg(arg_reg));
         }
 
-        assert!(return_type.is_unit());
         match *call_type {
             CallType::Ctor(_, _) => {
                 self.gen
--- a/dora/src/bytecode/generator.rs
+++ b/dora/src/bytecode/generator.rs
@@ -492,21 +492,9 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
         let call_type = self.src.map_calls.get(expr.id).unwrap().clone();
         let fct_id = call_type.fct_id().unwrap();
 
-        let fct = self.vm.fcts.idx(fct_id);
-        let fct = fct.read();
-
-        let callee_id = if fct.kind.is_definition() {
-            unimplemented!()
-        } else {
-            fct_id
-        };
-
+        let callee_id = fct_id;
         let callee = self.vm.fcts.idx(callee_id);
         let callee = callee.read();
-
-        if let FctKind::Builtin(_intrinsic) = callee.kind {
-            unimplemented!()
-        }
 
         let return_type = if dest.is_effect() {
             BuiltinType::Unit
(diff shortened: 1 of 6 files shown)


dora/src/bytecode/writer.rs
```diff
@@ -888,6 +968,24 @@ impl BytecodeWriter {
         self.emit_values(&values);
     }
 
+    fn emit_store_field(
```
dinfuehr: I think this is the same as `emit_load_field`, I think we could rename this function to e.g. `emit_field` and then use it for both stores and loads.

dora/src/cannon/codegen.rs
```diff
@@ -938,6 +990,191 @@ where
...
+
+    fn emit_new_object(&mut self, dest: Register, class_def_id: ClassDefId) {
+        assert_eq!(self.bytecode.register_type(dest), BytecodeType::Ptr);
+
+        let cls = self.vm.class_defs.idx(class_def_id);
+        let cls = cls.read();
+
+        self.asm.emit_comment(Comment::Alloc(class_def_id));
+
+        let alloc_size = match cls.size {
```
dinfuehr: I think we should always have fixed-size allocations in NewObject (we need an additional bytecode for allocation arrays).

dora/src/cannon/codegen.rs
```diff
@@ -938,6 +990,191 @@ where
...
+            InstanceSize::Fixed(size) => AllocationSize::Fixed(size as usize),
+            _ => unimplemented!(
+                "class size type {:?} for new object not supported",
+                cls.size
+            ),
+        };
+
+        let array_ref = match cls.size {
```
dinfuehr: This should always be false then.

dora/src/cannon/codegen.rs
```diff
@@ -938,6 +990,191 @@ where
...
+            InstanceSize::ObjArray => true,
+            _ => false,
+        };
+
+        let gcpoint = GcPoint::from_offsets(self.references.clone());
```
dinfuehr: Right now we need to pass a GcPoint in here, with Cannon this is probably not needed anymore. We might want to change this later.

dora/src/cannon/codegen.rs
```diff
@@ -938,6 +990,191 @@ where
...
+        // clear mark/fwdptr word in header
+        assert!(Header::size() == 2 * mem::ptr_width());
+        self.asm.load_int_const(MachineMode::Ptr, REG_TMP1, 0);
+        self.asm.store_mem(
+            MachineMode::Ptr,
+            Mem::Base(REG_RESULT, mem::ptr_width()),
+            REG_TMP1.into(),
+        );
+
+        match cls.size {
```
dinfuehr: Should always be `Fixed`

dora/src/bytecode/generator.rs
```diff
@@ -488,6 +488,77 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
+            .collect::<Vec<BytecodeType>>();
+        let num_args = arg_types.len();
+
+        let return_reg = if return_type.is_unit() {
+            Register::invalid()
+        } else {
+            self.ensure_register(dest, return_type.into())
+        };
+
+        let start_reg = if num_args > 0 {
```
dinfuehr: There should always be at least one register, right? Since self is always passed as first argument.

dinfuehr: (At least here in delegation)

dora/src/bytecode/generator.rs
```diff
@@ -488,6 +488,77 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
+            self.gen.add_register_chain(&arg_types)
+        } else {
+            Register::zero()
+        };
+
+        let arg_start_reg = if callee.has_self() {
```
dinfuehr: Here as well

dora/src/bytecode/generator.rs
```diff
@@ -488,6 +488,77 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
+        };
+
+        for (idx, arg) in expr.args.iter().enumerate() {
+            let arg_reg = arg_start_reg.offset(idx);
+            self.visit_expr(arg, DataDest::Reg(arg_reg));
+        }
+
+        match *call_type {
+            CallType::Ctor(_, _) => {
+                self.gen
```
dinfuehr: I would add an `assert(return_type.is_unit())` here. Otherwise this code might silently generate wrong in the future.

dora/src/bytecode/generator.rs
```diff
@@ -488,6 +488,77 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
 
+    fn visit_expr_delegation(&mut self, expr: &ExprDelegationType, dest: DataDest) -> Register {
+        let call_type = self.src.map_calls.get(expr.id).unwrap().clone();
+        let fct_id = call_type.fct_id().unwrap();
+
+        let fct = self.vm.fcts.idx(fct_id);
+        let fct = fct.read();
+
+        let callee_id = if fct.kind.is_definition() {
+            unimplemented!()
```
dinfuehr: IIRC this should not be true for delegation.

dora/src/bytecode/generator.rs
```diff
@@ -488,6 +488,77 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
+        } else {
+            fct_id
+        };
+
+        let callee = self.vm.fcts.idx(callee_id);
+        let callee = callee.read();
+
+        if let FctKind::Builtin(_intrinsic) = callee.kind {
```
dinfuehr: This neither.

tests/cannon/ctor1.dora
```diff
@@ -0,0 +1,12 @@
+//= output "xyz"
+
+@cannon class X(let msg: String) {
```
dinfuehr: Does this test even run the new delegation code? AFAIU there should only be a delegation-call when the class has a super class.

dora/src/bytecode/writer.rs
hzahradnik: I will remove the duplicate, but maybe we should be a little bit more specific for the name? What do you think about `emit_field_access` (or `emit_access_field`)?

dinfuehr: Ok

dora/src/cannon/codegen.rs
hzahradnik: Changed the `unimplemented` in the catchall arm to `unreachable`. An additional assert is not necessary.

dora/src/cannon/codegen.rs
hzahradnik: Removed.

dora/src/cannon/codegen.rs
hzahradnik: Removed the match and instead use `alloc_size` from above.

hzahradnik: Added a unreachable, but I guess there is a better way. I will look into this.

dora/src/bytecode/generator.rs
hzahradnik: Should i add an assertion to guarantee our assumption?

dora/src/bytecode/generator.rs
hzahradnik: Same as above: assertion to guarantee the assumption?

dora/src/bytecode/generator.rs
hzahradnik: This means implicitly, that `ensure_register` a few lines above is incorrect.

dora/src/bytecode/generator.rs
hzahradnik: I removed the code.

dora/src/bytecode/generator.rs
hzahradnik: This too.

tests/cannon/ctor1.dora
hzahradnik: You're correct, it did not. I added another test, which should rest delegation.

dora/src/cannon/codegen.rs
dinfuehr: Usually I add methods to the struct for that (here it would be `to_fixed()`)

dora/src/bytecode/generator.rs
dinfuehr: Why? AFAICS ensure_register can handle all types, only this code would silently generate the wrong bytecode when the result is actually needed.

dora/src/cannon/codegen.rs
```diff
@@ -938,6 +990,191 @@ where
...
+            REG_RESULT,
+            Position { line: 0, column: 0 }, // Correct position is currently not known.
+        );
+    }
+
+    fn emit_invoke_direct_void(&mut self, fct_id: FctId, start_reg: Register, num: u32) {
+        assert_eq!(self.bytecode.register_type(start_reg), BytecodeType::Ptr);
+
+        let _fct = self.vm.fcts.idx(fct_id);
+        let _fct = _fct.read();
```
dinfuehr: Here you could add an assert that the function does not have type params.

dora/src/bytecode/generator.rs
dinfuehr: Oh wait, yes you are right. Constructors should always return void, so the `ensure_register` above is not needed here.

dora/src/bytecode/generator.rs
```diff
@@ -488,6 +488,53 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
 
+    fn visit_expr_delegation(&mut self, expr: &ExprDelegationType, dest: DataDest) -> Register {
+        let call_type = self.src.map_calls.get(expr.id).unwrap().clone();
+        let fct_id = call_type.fct_id().unwrap();
+
+        let callee_id = fct_id;
+        let callee = self.vm.fcts.idx(callee_id);
+        let callee = callee.read();
+
+        let return_type = if dest.is_effect() {
```
dinfuehr: Constructors do not return a value, so we can simply assert `callee.return_type.is_unit()` here instead

dora/src/bytecode/generator.rs
```diff
@@ -488,6 +488,53 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
         }
     }
 
+    fn visit_expr_delegation(&mut self, expr: &ExprDelegationType, dest: DataDest) -> Register {
+        let call_type = self.src.map_calls.get(expr.id).unwrap().clone();
+        let fct_id = call_type.fct_id().unwrap();
```
dinfuehr: Nit: I don't think we need `fct_id` here.

dora/src/cannon/codegen.rs
dinfuehr: But let's keep it this way in this PR

dora/src/bytecode/generator.rs
hzahradnik: I already added the assert, but I missed to refactor this. Should be fixed.

dora/src/bytecode/generator.rs
hzahradnik: How would this work?

dora/src/bytecode/generator.rs
hzahradnik: I added also an assertion for `dest.is_effect()` to be pedantic. cf. `visit_expr_assign`.

dora/src/bytecode/generator.rs
hzahradnik: How would this work exactly?

dora/src/cannon/codegen.rs
hzahradnik: An assert for the class type params would also make sense, wouldn't it?

dora/src/bytecode/generator.rs
dinfuehr: fct_id and callee_id are the same, so one of the variables is not needed.

dora/src/cannon/codegen.rs
dinfuehr: Yes

PR #128, opened 2020-01-04
Drop most usages of nil from std lib and tests

soc:
Kept files nil1.dora - nil9.dora for now, they can
be deleted when the feature is removed for good.


dinfuehr:
I've skimmed the PR and I am not convinced we can merge this as-is.

For example I don't think we should remove test cases for nil as long as we still have it. There is code commented out, HACK in the comments, etc. I am also curious how much using Option regresses the benchmarks. Also String::fromBytesPartOrNull should probably be using exception handling or be rewritten in Dora. It feels that there is too much going in this single PR while each of these points probably warrants its own discussion.

dinfuehr:
I think we should split up this PR and start e.g. with removing nil from HashMap.

soc:
Closing this – everything but the tests changes are merged now.

PR #129, opened 2020-01-05
Migrate String#parseInt->Int|throws to String#toInt->Option[Int] ...

soc:
... same with String#parseLong.

### DIFF
--- a/dora/src/stdlib.rs
+++ b/dora/src/stdlib.rs
@@ -118,15 +118,19 @@ pub extern "C" fn sleep(seconds: i32) {
     thread::sleep(Duration::from_secs(seconds as u64));
 }
 
-pub extern "C" fn throw_native(val: bool) {
-    if val {
-        handle_scope(|| {
-            let vm = get_vm();
-            let obj = alloc_exception(vm, Ref::null());
-            let obj = root(obj);
+pub extern "C" fn throw_native() {
+    handle_scope(|| {
+        let vm = get_vm();
+        let obj = alloc_exception(vm, Ref::null());
+        let obj = root(obj);
 
-            exception_set(obj.direct().address())
-        })
+        exception_set(obj.direct().address())
+    })
+}
+
+pub extern "C" fn test_throw_native(val: bool) {
+    if val {
+        throw_native()
     }
 }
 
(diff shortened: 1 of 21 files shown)


tests/parse1.dora
```diff
@@ -1,6 +1,6 @@
 fun main() {
-  assert("1".parseInt() == 1);
-  assert("100".parseInt() == 100);
-  assert("789".parseInt() == 789);
-  assert("bla".parseInt() == 0);
+  assert("1".toInt().unwrap() == 1);
+  assert("100".toInt().unwrap() == 100);
+  assert("789".toInt().unwrap() == 789);
+  assert("bla".toInt().isNone());
```
soc: This doesn't work:

```
#> target/debug/dora tests/parse1.dora
uncaught exception
3: String.toIntOrThrow() -> Int: 27
2: String.toInt() -> Option[Int]: 30
1: main(): 5
```

Any idea why?

dora/src/stdlib.rs
```diff
@@ -249,18 +249,28 @@ pub extern "C" fn argv(ind: i32) -> Ref<Str> {
...
     let slice = val.content();
     let val = str::from_utf8(slice).unwrap();
 
-    val.parse::<i32>().unwrap_or(0)
+    let result = val.parse::<i32>();
+    if result.is_ok() {
+        return result.unwrap()
+    }
+    throw_native(true);
+    0
```
soc: Can't use `unreachable!()`, this is actually reached!?

tests/parse1.dora
dinfuehr: I think `toIntOrThrow` and `toLongOrThrow` need to be marked as `throws`. I would assume that this is the problem here, could also be a bug though.

dora/src/stdlib.rs
dinfuehr: Yes, throwing an exception from native code just sets a flag. When returning back to Dora code, there is assembly code that checks this flag and starts the exception handling process if it is set.

dora/src/stdlib.rs
```diff
@@ -249,18 +249,28 @@ pub extern "C" fn argv(ind: i32) -> Ref<Str> {
...
+pub extern "C" fn str_to_long_or_throw(val: Ref<Str>) -> i64 {
     let slice = val.content();
     let val = str::from_utf8(slice).unwrap();
 
-    val.parse::<i64>().unwrap_or(0)
+    let result = val.parse::<i64>();
+    if result.is_ok() {
+        return result.unwrap()
+    }
+    throw_native(true);
```
dinfuehr: `throw_native` is just a test function and isn't intended to be used apart from that. The boolean argument is just there to simulate a function that sometimes throws but also succeeds. Could you please create another function that just creates the exception object and sets the flag. throw_native could also use this function when the argument is true.

dinfuehr:
Very cool, thanks! I think adding `throws` to the function definition should fix the error, then we can merge this. (`self.toIntOrThrow()` needs to be invoked as `try self.toIntOrThrow()` then as well)

tests/parse1.dora
soc: Done, thanks!

dora/src/stdlib.rs
soc: Done!

soc:
@dinfuehr Updated!

Issue #117, opened 2019-12-25
Implement additional bit twiddling intrinsics


soc:
(This is probably a good beginner issue.)

There are a few operations that operate on integers that can be implemented manually, but could profit a lot from directly using the CPU's specially-built instruction for it.
All operations listed here have a direct counterpart in assembly.

The goal is to provide a method for these operations that make use of the special CPU instruction:

- [x] [POPCNT](https://www.felixcloutier.com/x86/popcnt) — Return the Count of Number of Bits Set to 1 (https://github.com/dinfuehr/dora/pull/131)
- [x] [LZCNT](https://www.felixcloutier.com/x86/lzcnt) — Count the Number of Leading Zero Bits (https://github.com/dinfuehr/dora/pull/131)
- [x] [TZCNT](https://www.felixcloutier.com/x86/tzcnt) — Count the Number of Trailing Zero Bits (https://github.com/dinfuehr/dora/pull/131)
- [x] [ROL, ROR](https://www.felixcloutier.com/x86/rcl:rcr:rol:ror) — Rotate Left/Rotate Right (https://github.com/dinfuehr/dora/pull/138)
- [x] [SHLX, SHRX, SARX](https://www.felixcloutier.com/x86/sarx:shlx:shrx) — Shift without doing stupid shit (https://github.com/dinfuehr/dora/pull/144)

For each of these instructions the following steps can be taken:

1. Add the method to `Int.dora`/`Long.dora` and mark it as `@internal`.
2. Add the method to the list of intrinsic functions in `prelude.rs`.
3. From there, you can use existing functions as an example of where and how to wire things up in `codegen.rs`.
4. Write tests to verify that the function works as expected. Make sure to identify and cover corner cases as well!

---

[Bit Manipulation Instruction Sets](https://en.wikipedia.org/wiki/Bit_Manipulation_Instruction_Sets) gives a good overview of additional operations that could be of interest.

soc:
Further thoughts:

- Would it make sense to also add functions for the (trivial) inverse operations ("Count of Number of Bits Set to 0", "Number of Leading One Bits", "Number of Trailing One Bits")?
- What's a good, consistent naming scheme for these functions? Ideas:
  - `countZeroes/countOnes`, `countLeadingZeroes/countLeadingOnes`, `countTrailingZeroes/countTrailingOnes`
[...]

PR #131, opened 2020-01-07
Implement POPCNT/LZCNT/TZCNT

soc:
Implements the three examples from #117.


soc:
I worked on most of this stuff on a machine without those instructions, so it still needs fixes and testing etc.

dinfuehr:
Already looks really good 👍 There can't miss much, I can help you out here - although probably not before the weekend.

soc:
Cool, thanks!

Status:
- Trailing should be working (except I haven't tested it on a machine that supports it.
- Leading doesn't work for reasons I don't understand (wrong values).
- Popcount crashes the VM with trap 4 (no idea why).

soc:
Output from int3.dora indicates that it's running `BSR` instead of `LZCNT` (it returns the index, not the count) – despite running on hardware that's new enough to support `LZCNT` (Intel(R) Xeon(R) CPU E3-1220 V2) ... I don't get it.
```
=== Count Zero Bits Leading ===
1 00000000000000000000000000000001 0
2 00000000000000000000000000000010 1
35 00000000000000000000000000100011 5
39 00000000000000000000000000100111 5
1024 00000000000000000000010000000000 10
Int::max_value() 01111111111111111111111111111111 30
-1 11111111111111111111111111111111 31
Int::min_value() 10000000000000000000000000000000 31
=== Count One Bits Leading ===
0 00000000000000000000000000000000 31
1 00000000000000000000000000000001 31
2 00000000000000000000000000000010 31
35 00000000000000000000000000100011 31
39 00000000000000000000000000100111 31
1024 00000000000000000000010000000000 31
Int::max_value() 01111111111111111111111111111111 31
Int::min_value() 10000000000000000000000000000000 30
```

dora/src/masm/x64.rs
```diff
@@ -499,6 +499,64 @@ impl MacroAssembler {
...
+            MachineMode::Int64 => 1,
+            _ => unimplemented!(),
+        };
+
+        if count_one_bits {
+            asm::popcnt(self, x64, dest, src);
+        } else {
+            asm::popcnt(self, x64, REG_TMP1, src);
+            asm::emit_movl_imm_reg(self, 32, dest);
+            asm::emit_sub_reg_reg(self, x64, REG_TMP1, dest)
```
soc: Should probably use `asm::emit_not_reg` like the funs below ...

soc:
All variations of count ones/zeroes leading/trailing work for me; popcount still crashes.

dinfuehr:
I will try to take a look the next few days. Can you please rebase it to master?

I suppose there is something wrong with the encoding. Did you try --emit-asm to see whether it prints the instruction you intend?

soc:
> Can you please rebase it to master?

Will do!

> I suppose there is something wrong with the encoding. Did you try --emit-asm to see whether it prints the instruction you intend?

Ah, didn't remember that option, thanks! I will try it.

soc:
It works fine on a machine that is recent enough; not sure about the build failure.

Is there a way to print the output of failing tests by default?

soc:
@dinfuehr I get the same `test returned 105` failure on a machine that doesn't support the instructions – could it be that the macOS machine's CPU used for CI is just old?

dinfuehr:
Works on my Macbook. `tester.rb` doesn't support this right now, but you can just add a puts statement into the `run_test` method to print stdout/stderr.

I fear that we might have to introduce some form of check and fallback for this...

dora/src/semck/prelude.rs
```diff
@@ -439,7 +456,7 @@ fn internal_method<'ast>(vm: &mut VM<'ast>, clsid: ClassId, name: &str, kind: Fc
         let mtd = vm.fcts.idx(mid);
         let mut mtd = mtd.write();
 
-        if mtd.name == name && mtd.internal {
+        if mtd.name == name {
             mtd.kind = kind;
             mtd.internal_resolved = true;
```
soc: Should we set `mtd.internal = true` here too?

dora/src/semck/prelude.rs
```diff
@@ -144,6 +145,12 @@ fn find_trait<'ast>(vm: &mut VM<'ast>, name: &str) -> TraitId {
 }
 
 pub fn internal_functions<'ast>(vm: &mut VM<'ast>) {
+    let cpuid = CpuId::new();
```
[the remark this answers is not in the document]
soc: Ah, right, good point! Will do!

dora/Cargo.toml
```diff
@@ -20,6 +20,7 @@ dora-parser = { path = "../dora-parser", version = "0.0.2" }
 threadpool = "1.7.*"
 crossbeam-deque = "0.7.*"
 rand = "0.7.*"
+raw-cpuid = "7.0.*"
```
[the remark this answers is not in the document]
soc: Yes, will do!

dora/src/semck/prelude.rs
```diff
@@ -259,6 +300,40 @@ pub fn internal_functions<'ast>(vm: &mut VM<'ast>) {
     intrinsic_method(vm, clsid, "unaryMinus", Intrinsic::LongNeg);
     intrinsic_method(vm, clsid, "not", Intrinsic::LongNot);
 
+    if has_popcnt {
+        intrinsic_method(vm, clsid, "countZeroBits", Intrinsic::LongCountZeroBits);
+        intrinsic_method(vm, clsid, "countOneBits", Intrinsic::LongCountOneBits);
+    }
+    // has_lzcount
```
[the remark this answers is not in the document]
soc: The condition could be weakened from "has_tzcount" to "has_lzcount", but the cpuid library lacks support for checking that flag. See https://github.com/gz/rust-cpuid/issues/30.

Will add a few words here.

dora/src/semck/prelude.rs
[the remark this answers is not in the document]
soc: https://github.com/dinfuehr/dora/pull/139

dora/src/semck/prelude.rs
soc: Wouldn't it be cleaner to run `typeck::check` before we are adding the internal methods?

soc:
@dinfuehr Do you think this is good to go as it is now?

dora/stdlib/Int.dora
```diff
@@ -42,6 +42,43 @@
   @internal fun unaryMinus() -> Int;
   @internal fun not() -> Int;
 
+  @internal fun countZeroBits() -> Int = self.not().countOneBits();
+  @internal fun countOneBits() -> Int {
```
soc: The change in internalck now allows marking methods and functions with implementations as internal.

dora/src/semck.rs
```diff
@@ -155,7 +156,7 @@ fn internalck<'ast>(vm: &VM<'ast>) {
             let method = vm.fcts.idx(*method);
             let method = method.read();
 
-            if method.internal && !method.internal_resolved {
+            if method.internal && !method.internal_resolved && method.kind.is_definition() {
```
soc: This change means that having `@internal`, but not having decided to use an intrinsic is only considered to be a problem if we don't have an implementation in the source code.

soc:
@dinfuehr is this good to go?

dora/src/semck/prelude.rs
dinfuehr: I am still not convinced that replacing `kind` is the way to go - but since I can't name an example where we need it in the runtime, we can do this.

dora/src/semck.rs
```diff
@@ -93,14 +93,15 @@ pub fn check<'ast>(vm: &mut VM<'ast>) {
...
-    // define internal functions & methods
-    prelude::internal_functions(vm);
     prelude::known_methods(vm);
 
     // check types of expressions in functions
     typeck::check(vm);
     return_on_error!(vm);
 
+    // define internal functions & methods
+    prelude::internal_functions(vm);
```
dinfuehr: This is called too early, this prevents `flowck` and `returnck` from happening on such functions. I wouldn't move that function after those two calls but also out of `semck`. I would rather add a function `prelude::install_conditional_intrinsics` that we can invoke after `semck::check` in `driver/start.rs`. This will ensure that we do all the semantic checks in `semck`, also this will make a refactor easier that I plan to do (I want to move the semantic analysis to a separate library - for that the `semck` needs to run without the `VM` but with its own data structures, so that we only create a `VM` with a valid program.).

soc: Fixed in second commit (feel free to squash after review if you are happy with it).

dinfuehr:
Thanks!

soc:
Thank you too!

PR #137, opened 2020-01-24
Various naming cleanups

soc:
- max/min values got aligned with the general naming scheme
- tried to make right shifts a bit more consistent (see commit message).

@dinfuehr What do you think?


dinfuehr:
Nice cleanup!

PR #138, opened 2020-01-25
Implement {Int,Long}#{rotateLeft,rotateRight}

soc:
More operations for https://github.com/dinfuehr/dora/issues/117.

### DIFF
--- a/dora/src/masm/x64.rs
+++ b/dora/src/masm/x64.rs
@@ -474,6 +474,44 @@ impl MacroAssembler {
         }
     }
 
+    pub fn int_rol(&mut self, mode: MachineMode, dest: Reg, lhs: Reg, rhs: Reg) {
+        let x64 = match mode {
+            MachineMode::Int32 => 0,
+            MachineMode::Int64 => 1,
+            _ => unimplemented!(),
+        };
+
+        if rhs != RCX {
+            assert!(lhs != RCX);
+            asm::emit_mov_reg_reg(self, x64, rhs, RCX);
+        }
+
+        asm::emit_rol_reg_cl(self, x64, lhs);
+
+        if dest != lhs {
+            asm::emit_mov_reg_reg(self, x64, lhs, dest);
+        }
+    }
+
+    pub fn int_ror(&mut self, mode: MachineMode, dest: Reg, lhs: Reg, rhs: Reg) {
+        let x64 = match mode {
+            MachineMode::Int32 => 0,
+            MachineMode::Int64 => 1,
+            _ => unimplemented!(),
+        };
+
+        if rhs != RCX {
+            assert!(lhs != RCX);
+            asm::emit_mov_reg_reg(self, x64, rhs, RCX);
+        }
+
+        asm::emit_ror_reg_cl(self, x64, lhs);
+
+        if dest != lhs {
+            asm::emit_mov_reg_reg(self, x64, lhs, dest);
+        }
+    }
+
     pub fn int_or(&mut self, mode: MachineMode, dest: Reg, lhs: Reg, rhs: Reg) {
         let x64 = match mode {
             MachineMode::Int32 => 0,
--- a/dora/src/baseline/codegen.rs
+++ b/dora/src/baseline/codegen.rs
@@ -1993,6 +1993,13 @@ where
                 self.emit_intrinsic_bin_call(args[0], args[1], dest, intrinsic, pos)
             }
 
+            Intrinsic::IntRotateLeft => {
+                self.emit_intrinsic_bin_call(args[0], args[1], dest, intrinsic, pos)
+            }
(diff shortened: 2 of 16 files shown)


dinfuehr:
The rotate changes look good to me. Could you add some more test case for what happens if you rotate more than 31 or 63 bits? At least with shifting there is a difference between some ISAs. Would be good to have the expected behavior tested.

I am also wondering about the right type for the shift/rotate argument. Currently it is Long for Long, but not sure this makes sense because you can't shift/rotate more than 63bits anyway (which fits into Int or even Byte). So we actually don't need the full range. I would merge it as-is, but this might be something to think about.

soc:
> The rotate changes look good to me. Could you add some more test case for what happens if you rotate more than 31 or 63 bits? At least with shifting there is a difference between some ISAs. Would be good to have the expected behavior tested.

Agreed, will do this!

> I am also wondering about the right type for the shift/rotate argument. Currently it is Long for Long, but not sure this makes sense because you can't shift/rotate more than 63bits anyway (which fits into Int or even Byte). So we actually don't need the full range. I would merge it as-is, but this might be something to think about.

I was planning to propose changing all shift-by/rotate-by to `Int`s after this change, but I can also add another commit if we agree on this.

soc:
Regarding CPUID checks – maybe it makes sense to split the CPUID stuff from the popcount PR and get that ready independently. What do you think?

dinfuehr:
> I was planning to propose changing all shift-by/rotate-by to Ints after this change, but I can also add another commit if we agree on this.
You don't need to change this in this PR - unless you want to of course. Whatever makes sense to you.

> Regarding CPUID checks – maybe it makes sense to split the CPUID stuff from the popcount PR and get that ready independently. What do you think?
A separate PR might make sense indeed, but choose the approach you prefer.

dinfuehr:
> I'll consult the Intel and ARM specs whether they differ in behavior in this case, and if yes, if one of them makes more sense.

At least for shifting the behavior was changed from ARMv7 to ARMv8 IIRC. They switched to the intel behavior. So I think we should pick the Intel one.

soc:
@dinfuehr I added commits for the shift/rotate argument type and tests.

dora/src/masm/x64.rs
```diff
@@ -474,6 +474,44 @@ impl MacroAssembler {
         }
     }
 
+    pub fn int_rol(&mut self, mode: MachineMode, dest: Reg, lhs: Reg, rhs: Reg) {
+        let x64 = match mode {
+            MachineMode::Int32 => 0,
+            MachineMode::Int64 => 1,
+            _ => unimplemented!(),
+        };
+
```
dinfuehr: Shall we assert here that `lhs` should not be `RCX`?

dora/src/masm/x64.rs
```diff
@@ -474,6 +474,44 @@ impl MacroAssembler {
...
+        }
+    }
+
+    pub fn int_ror(&mut self, mode: MachineMode, dest: Reg, lhs: Reg, rhs: Reg) {
+        let x64 = match mode {
+            MachineMode::Int32 => 0,
+            MachineMode::Int64 => 1,
+            _ => unimplemented!(),
+        };
+
```
dinfuehr: Here again

dora/src/masm/x64.rs
soc: Yes, good point!

soc:
Yes, absolutely! Do we have any other instructions like this were it might make sense to check that we don't clobber registers?

(My plan after having CPUID checks is to replace shift and rotate instructions with the newer, less broken ones where available, but as we also probably want to support the older ones, the check should be done.)

dinfuehr:
Not sure what exactly you mean with broken, but probably division.

soc:
The variations without the `X` affect flags all over the place, are slower and have these crazy rules from where arguments are read. E. g. compare [old](https://www.felixcloutier.com/x86/sal:sar:shl:shr) with [new](https://www.felixcloutier.com/x86/sarx:shlx:shrx). Sadly, they require BMI2.

I'll check division needs such an assert, too!

dora/src/masm/x64.rs
soc: Mhhh, aren't we doing this two lines below?

dinfuehr: Oh right, missed that.... Sorry...

soc: No problem! So this is good to go, right?

dinfuehr:
That they are updating flags is indeed super surprising.... Since the assert was in, this PR is finished as well, right?

soc:
@dinfuehr Yes! Everything else is probably a follow-up commit!

dinfuehr:
Merged, thanks!

PR #139, opened 2020-01-25
Add support for CPU feature checks

soc:
Functions return `true` on ARM, since the expected instructions have been available for more than 15 years.

### DIFF
--- a/dora/src/cpu/x64.rs
+++ b/dora/src/cpu/x64.rs
@@ -2,6 +2,8 @@ use std::sync::atomic::{compiler_fence, Ordering};
 
 pub use self::param::*;
 pub use self::reg::*;
+use lazy_static::lazy_static;
+use raw_cpuid::{CpuId, ExtendedFeatures, FeatureInfo};
 
 pub mod asm;
 pub mod param;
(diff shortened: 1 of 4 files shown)


dora/src/cpu/x64.rs
```diff
@@ -11,3 +13,27 @@ pub fn flush_icache(_: *const u8, _: usize) {
...
+}
+
+lazy_static! {
+static ref FEATURES: FeatureInfo = CpuId::new().get_feature_info().unwrap();
+static ref FEATURES_EXTENDED: ExtendedFeatures = CpuId::new().get_extended_feature_info().unwrap();
+
+static ref HAS_POPCNT: bool = FEATURES.has_popcnt();
+static ref HAS_TZCNT: bool = FEATURES_EXTENDED.has_bmi1();
+// support for MULX, RORX, SARX, SHRX, SHLX
+static ref HAS_X_OPS: bool = FEATURES_EXTENDED.has_bmi2();
```
soc: This field has no `pub` function, because both the "bad old" instructions as well as the possible "better" replacements are intrinsic, so we don't need to make this decision from the outside.

PR #143, opened 2020-01-27
Convert HashMap from nil to Option

soc:
As suggested in https://github.com/dinfuehr/dora/pull/128#issuecomment-571072355.


dinfuehr:
Cool, thanks!

PR #144, opened 2020-01-28
Use shift instructions that don't set flags where available

soc:
Follow-up to https://github.com/dinfuehr/dora/pull/138.

### DIFF
--- a/dora/src/cpu/x64/asm.rs
+++ b/dora/src/cpu/x64/asm.rs
@@ -2,6 +2,19 @@ use crate::cpu::*;
 use crate::masm::{CondCode, Label, MacroAssembler};
 use crate::ty::MachineMode;
 
+pub const VEXM_0F: u8 = 1;
+pub const VEXM_0F38: u8 = 2;
+pub const VEXM_0F3A: u8 = 3;
+
+pub const VEXL_Z: u8 = 0;
+pub const VEXL_128: u8 = 0;
+pub const VEXL_256: u8 = 1;
+
+pub const VEXP_NONE: u8 = 0;
+pub const VEXP_66: u8 = 1;
+pub const VEXP_F3: u8 = 2;
+pub const VEXP_F2: u8 = 3;
+
 pub fn emit_or_reg_reg(buf: &mut MacroAssembler, x64: u8, src: Reg, dest: Reg) {
     emit_alu_reg_reg(buf, x64, 0x09, src, dest);
 }
@@ -376,13 +389,42 @@ pub fn emit_op(buf: &mut MacroAssembler, opcode: u8) {
     buf.emit_u8(opcode);
 }
 
+pub fn emit_vex3_rxbm(buf: &mut MacroAssembler, r: u8, x: u8, b: u8, m: u8) {
+    assert!(r == 0 || r == 1);
+    assert!(x == 0 || x == 1);
+    assert!(b == 0 || b == 1);
+    assert!(m > 1 && m < 4);
+
+    buf.emit_u8(0b11000100);
+    buf.emit_u8(!r << 7 | (!x << 6) & 0b1000000 | (!b << 5) & 0b100000 | m);
+}
+
+pub fn emit_vex3_wvlp(buf: &mut MacroAssembler, w: u8, v: u8, l: u8, p: u8) {
+    assert!(w == 0 || w == 1);
+    assert!(v < 16);
+    assert!(l == 0 || l == 1);
+    assert!(p < 4);
+
+    buf.emit_u8(w << 7 | (!v << 3) & 0b1111000 | l << 2 | p);
+}
+
+pub fn emit_vex2(buf: &mut MacroAssembler, r: u8, v: u8, l: u8, p: u8) {
+    assert!(r == 0 || r == 1);
+    assert!(v < 16);
+    assert!(l == 0 || l == 1);
+    assert!(p < 4);
+
+    buf.emit_u8(0b11000101);
+    buf.emit_u8(!r << 7 | (!v << 3) & 0b1111000 | l << 2 | p);
+}
+
 pub fn emit_rex(buf: &mut MacroAssembler, w: u8, r: u8, x: u8, b: u8) {
     assert!(w == 0 || w == 1);
     assert!(r == 0 || r == 1);
     assert!(x == 0 || x == 1);
     assert!(b == 0 || b == 1);
 
-    buf.emit_u8(0x4 << 4 | w << 3 | r << 2 | x << 1 | b);
+    buf.emit_u8(0x40 | w << 3 | r << 2 | x << 1 | b);
 }
 
 pub fn emit_modrm(buf: &mut MacroAssembler, mode: u8, reg: u8, rm: u8) {
@@ -858,6 +900,13 @@ pub fn emit_callq_reg(buf: &mut MacroAssembler, dest: Reg) {
     emit_modrm(buf, 0b11, 0b10, dest.and7());
 }
 
+pub fn emit_shlx(buf: &mut MacroAssembler, x64: u8, dest: Reg, lhs: Reg, rhs: Reg) {
+    emit_vex3_rxbm(buf, dest.msb(), 0, lhs.msb(), VEXM_0F38);
+    emit_vex3_wvlp(buf, x64, rhs.int(), VEXL_Z, VEXP_66);
+    emit_u8(buf, 0xF7);
+    emit_modrm(buf, 0b11, dest.and7(), lhs.and7());
+}
+
 pub fn emit_shlq_reg(buf: &mut MacroAssembler, imm: u8, dest: Reg) {
     emit_rex(buf, 1, 0, 0, dest.msb());
     emit_op(buf, 0xC1);
@@ -884,6 +933,13 @@ pub fn emit_shl_reg_cl(buf: &mut MacroAssembler, x64: u8, dest: Reg) {
     emit_modrm(buf, 0b11, 0b100, dest.and7());
 }
 
+pub fn emit_shrx(buf: &mut MacroAssembler, x64: u8, dest: Reg, lhs: Reg, rhs: Reg) {
+    emit_vex3_rxbm(buf, dest.msb(), 0, lhs.msb(), VEXM_0F38);
+    emit_vex3_wvlp(buf, x64, rhs.int(), VEXL_Z, VEXP_F2);
+    emit_u8(buf, 0xF7);
+    emit_modrm(buf, 0b11, dest.and7(), lhs.and7());
+}
+
 pub fn emit_shr_reg_cl(buf: &mut MacroAssembler, x64: u8, dest: Reg) {
     if dest.msb() != 0 || x64 != 0 {
         emit_rex(buf, x64, 0, 0, dest.msb());
@@ -904,6 +960,13 @@ pub fn emit_shr_reg_imm(buf: &mut MacroAssembler, x64: u8, dest: Reg, imm: u8) {
     if imm != 1 {
         emit_u8(buf, imm);
     }
+}
+
+pub fn emit_sarx(buf: &mut MacroAssembler, x64: u8, dest: Reg, lhs: Reg, rhs: Reg) {
+    emit_vex3_rxbm(buf, dest.msb(), 0, lhs.msb(), VEXM_0F38);
+    emit_vex3_wvlp(buf, x64, rhs.int(), VEXL_Z, VEXP_F3);
+    emit_u8(buf, 0xF7);
+    emit_modrm(buf, 0b11, dest.and7(), lhs.and7());
 }
 
 pub fn emit_sar_reg_cl(buf: &mut MacroAssembler, x64: u8, dest: Reg) {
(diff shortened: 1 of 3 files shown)


dora/src/cpu/x64/reg.rs
```diff
@@ -67,6 +67,19 @@ pub const XMM13: FReg = FReg(13);
...
+pub const VEXM_0F3A: u8 = 3;
+
+pub const VEXL_Z: u8 = 0;
+pub const VEXL_128: u8 = 0;
+pub const VEXL_256: u8 = 1;
+
+pub const VEXP_NONE: u8 = 0;
+pub const VEXP_66: u8 = 1;
+pub const VEXP_F3: u8 = 2;
+pub const VEXP_F2: u8 = 3;
```
soc: Is there a more appropriate place for these constants?

dora/src/masm/x64.rs
```diff
@@ -417,6 +418,19 @@ impl MacroAssembler {
         }
     }
 
+    pub fn int_shlx(&mut self, mode: MachineMode, dest: Reg, lhs: Reg, rhs: Reg) {
+        let x64 = match mode {
+            MachineMode::Int32 => 0,
+            MachineMode::Int64 => 1,
+            _ => unimplemented!(),
+        };
```
soc: (Not relevant for this commit.) Would it make sense in general to turn the `x64` snippet into a function instead of duplicating it in many places?

soc:
Need to check the failures on a system that supports those instructions. :-)

dora/src/compiler/asm.rs
```diff
@@ -263,15 +265,27 @@ where
     }
 
     pub fn int_shl(&mut self, mode: MachineMode, dest: Reg, lhs: Reg, rhs: Reg) {
-        self.masm.int_shl(mode, dest, lhs, rhs);
+        if has_x_ops() {
```
dinfuehr: This should probably move to `masm/x64.rs` since this is x64-specific. For example ARM64 doesn't have different shift operations AFAIK.

dora/src/cpu/x64/reg.rs
dinfuehr: They could just go into `masm/x64.rs` or `x64/asm.rs` since all the other code doesn't really care which shift operations are used.

dora/src/masm/x64.rs
```diff
@@ -417,6 +418,19 @@ impl MacroAssembler {
...
+
+        emit_vex3_rxbm(self, rhs.msb(), 0, lhs.msb(), VEXM_0F38);
```
dinfuehr: Implementation of these operations could go into `x64/asm.rs`. But I guess that's a matter of taste. I am trying to refactor the Assembler/MacroAssembler anyways. I am creating an Assembler in `dora/src/asm.rs` and I am going to move all CPU-specific instructions into that directory. So in the future this should be implemented in `dora/src/asm/x64.rs`.

dora/src/cpu/x64/reg.rs
dinfuehr: I would move this to `masm/x64.rs` or `x64/asm.rs`.

dora/src/masm/x64.rs
dinfuehr: Yes, I want to refactor the whole Assembler (see dora/src/asm.rs).

dora/src/compiler/asm.rs
soc: Ah yes, will do.

dora/src/cpu/x64/reg.rs
soc: Will do.

soc:
Turns out the result on CPUs with BMI2 support is either -1 or 0, depending on whether the input is even or odd. Need to investigate, not seeing any obvious mistakes...

dora/src/masm/x64.rs
soc: > Implementation of these operations could go into x64/asm.rs

Do you mean `emit_vex3_rxbm` etc. or the enclosing function, because the former already are.

soc:
@dinfuehr I think I addressed your comments, now I only need to figure out why the instructions don't work. :-)

dora/src/masm/x64.rs
dinfuehr: I was thinking about the enclosing function.

dora/src/cpu/x64/asm.rs
```diff
@@ -393,13 +393,42 @@ pub fn emit_op(buf: &mut MacroAssembler, opcode: u8) {
...
+}
+
 pub fn emit_rex(buf: &mut MacroAssembler, w: u8, r: u8, x: u8, b: u8) {
     assert!(w == 0 || w == 1);
     assert!(r == 0 || r == 1);
     assert!(x == 0 || x == 1);
     assert!(b == 0 || b == 1);
 
-    buf.emit_u8(0x4 << 4 | w << 3 | r << 2 | x << 1 | b);
+    buf.emit_u8(1 << 6 | w << 3 | r << 2 | x << 1 | b);
```
dinfuehr: I would propose to use `0x40` instead here, makes it easier to recognize this as rex-byte.

dora/src/masm/x64.rs
```diff
@@ -418,6 +432,27 @@ impl MacroAssembler {
...
+        emit_vex3_wvlp(self, x64, dest.int(), VEXL_Z, VEXP_66);
+        emit_u8(self, 0xF7);
+        emit_modrm(self, 0b11, rhs.and7(), lhs.and7());
+    }
+
+    pub fn int_shl_c(&mut self, mode: MachineMode, dest: Reg, lhs: Reg, rhs: Reg) {
```
dinfuehr: The idea behind the MacroAssembler is to provide a CPU-independent interface. It might not always be perfect but the functions `int_shl_c`, `int_shrx`, etc. shouldn't be public and/or be moved to x64/asm.rs.

dora/src/masm/x64.rs
soc: Ah, right, thanks!

soc:
Added second commit implementing your review feedback, if this looks ok I'll squash it.

dinfuehr:
The PR looks really good, thanks :+1:  The tests do not pass though.

soc:
Rebased on latest master – but there is still the problem with the asm not doing what it should do. :-/

@dinfuehr could you have a look? Maybe you are seeing something I missed.

soc:
@dinfuehr Shall we close this? I feel that if I haven't found the issue until now, I probably won't see it in the future either...

dinfuehr:
That would be really unfortunate since you've spent quite some time on it. Could you please rebase it? I will try to take a look, forgot about it unfortunately...

soc:
@dinfuehr Sure, will do!

soc:
@dinfuehr done!

dinfuehr:
I guess the problem is that you have swapped the arguments. The assembly tests don't cover this since you've always used the same registers for `dest` and `rhs`. When I swap those arguments all tests pass.

soc:
Mhhh, I tried:

```
Index: dora/src/cpu/x64/asm.rs
===================================================================
--- dora/src/cpu/x64/asm.rs	(revision 3c75de57240d3bbbd96630afe6bed00b56b14a95)
+++ dora/src/cpu/x64/asm.rs	(date 1582926959713)
@@ -901,8 +901,8 @@
 }
 
 pub fn emit_shlx(buf: &mut MacroAssembler, x64: u8, dest: Reg, lhs: Reg, rhs: Reg) {
-    emit_vex3_rxbm(buf, rhs.msb(), 0, lhs.msb(), VEXM_0F38);
-    emit_vex3_wvlp(buf, x64, dest.int(), VEXL_Z, VEXP_66);
+    emit_vex3_rxbm(buf, dest.msb(), 0, lhs.msb(), VEXM_0F38);
+    emit_vex3_wvlp(buf, x64, rhs.int(), VEXL_Z, VEXP_66);
     emit_u8(buf, 0xF7);
     emit_modrm(buf, 0b11, rhs.and7(), lhs.and7());
 }
@@ -934,8 +934,8 @@
 }
 
 pub fn emit_shrx(buf: &mut MacroAssembler, x64: u8, dest: Reg, lhs: Reg, rhs: Reg) {
-    emit_vex3_rxbm(buf, rhs.msb(), 0, lhs.msb(), VEXM_0F38);
-    emit_vex3_wvlp(buf, x64, dest.int(), VEXL_Z, VEXP_F2);
+    emit_vex3_rxbm(buf, dest.msb(), 0, lhs.msb(), VEXM_0F38);
+    emit_vex3_wvlp(buf, x64, rhs.int(), VEXL_Z, VEXP_F2);
     emit_u8(buf, 0xF7);
     emit_modrm(buf, 0b11, rhs.and7(), lhs.and7());
 }
@@ -963,8 +963,8 @@
 }
 
 pub fn emit_sarx(buf: &mut MacroAssembler, x64: u8, dest: Reg, lhs: Reg, rhs: Reg) {
-    emit_vex3_rxbm(buf, rhs.msb(), 0, lhs.msb(), VEXM_0F38);
-    emit_vex3_wvlp(buf, x64, dest.int(), VEXL_Z, VEXP_F3);
+    emit_vex3_rxbm(buf, dest.msb(), 0, lhs.msb(), VEXM_0F38);
+    emit_vex3_wvlp(buf, x64, rhs.int(), VEXL_Z, VEXP_F3);
     emit_u8(buf, 0xF7);
     emit_modrm(buf, 0b11, rhs.and7(), lhs.and7());
 }
```
but it fails as before.

dinfuehr:
You also need to change the `emit_vex3_rxbm` function call.

dinfuehr:
I think you should add some more test cases or at least change them, such that all arguments use different registers and also maybe different msb. You can get the encoding for those instructions easily with this [online assembler](https://defuse.ca/online-x86-assembler.htm)

dinfuehr:
I need to check whether I still have those changes. I fear that I didn't save those though, sorry for that... I think that the only other change I made was to replace `rhs` with `dest` in the emit_vex3_rxbm-invocations above.

dinfuehr:
That patch seems to fix all tests for me (at least on my machine):

```diff
diff --git a/dora/src/cpu/x64/asm.rs b/dora/src/cpu/x64/asm.rs
index d3f137a8..186c78b3 100644
--- a/dora/src/cpu/x64/asm.rs
+++ b/dora/src/cpu/x64/asm.rs
@@ -901,10 +901,10 @@ pub fn emit_callq_reg(buf: &mut MacroAssembler, dest: Reg) {
 }
 
 pub fn emit_shlx(buf: &mut MacroAssembler, x64: u8, dest: Reg, lhs: Reg, rhs: Reg) {
-    emit_vex3_rxbm(buf, rhs.msb(), 0, lhs.msb(), VEXM_0F38);
-    emit_vex3_wvlp(buf, x64, dest.int(), VEXL_Z, VEXP_66);
+    emit_vex3_rxbm(buf, dest.msb(), 0, lhs.msb(), VEXM_0F38);
+    emit_vex3_wvlp(buf, x64, rhs.int(), VEXL_Z, VEXP_66);
     emit_u8(buf, 0xF7);
-    emit_modrm(buf, 0b11, rhs.and7(), lhs.and7());
+    emit_modrm(buf, 0b11, dest.and7(), lhs.and7());
 }
 
 pub fn emit_shlq_reg(buf: &mut MacroAssembler, imm: u8, dest: Reg) {
@@ -934,10 +934,10 @@ pub fn emit_shl_reg_cl(buf: &mut MacroAssembler, x64: u8, dest: Reg) {
 }
 
 pub fn emit_shrx(buf: &mut MacroAssembler, x64: u8, dest: Reg, lhs: Reg, rhs: Reg) {
-    emit_vex3_rxbm(buf, rhs.msb(), 0, lhs.msb(), VEXM_0F38);
-    emit_vex3_wvlp(buf, x64, dest.int(), VEXL_Z, VEXP_F2);
+    emit_vex3_rxbm(buf, dest.msb(), 0, lhs.msb(), VEXM_0F38);
+    emit_vex3_wvlp(buf, x64, rhs.int(), VEXL_Z, VEXP_F2);
     emit_u8(buf, 0xF7);
-    emit_modrm(buf, 0b11, rhs.and7(), lhs.and7());
+    emit_modrm(buf, 0b11, dest.and7(), lhs.and7());
 }
 
 pub fn emit_shr_reg_cl(buf: &mut MacroAssembler, x64: u8, dest: Reg) {
@@ -963,10 +963,10 @@ pub fn emit_shr_reg_imm(buf: &mut MacroAssembler, x64: u8, dest: Reg, imm: u8) {
 }
 
 pub fn emit_sarx(buf: &mut MacroAssembler, x64: u8, dest: Reg, lhs: Reg, rhs: Reg) {
-    emit_vex3_rxbm(buf, rhs.msb(), 0, lhs.msb(), VEXM_0F38);
-    emit_vex3_wvlp(buf, x64, dest.int(), VEXL_Z, VEXP_F3);
+    emit_vex3_rxbm(buf, dest.msb(), 0, lhs.msb(), VEXM_0F38);
+    emit_vex3_wvlp(buf, x64, rhs.int(), VEXL_Z, VEXP_F3);
     emit_u8(buf, 0xF7);
-    emit_modrm(buf, 0b11, rhs.and7(), lhs.and7());
+    emit_modrm(buf, 0b11, dest.and7(), lhs.and7());
 }
 
 pub fn emit_sar_reg_cl(buf: &mut MacroAssembler, x64: u8, dest: Reg) {

```

soc:
Ah, I forgot the change to `emit_modrm` – `emit_vex3_rxbm` and `emit_vex3_wvlp` are as in my diff.

I will test and update my PR; thanks a lot for our help!

soc:
@dinfuehr Fixed and pushed. Sorry for forgetting about https://defuse.ca/online-x86-assembler.htm, would have shown me my mistakes quite easily, I guess. :-(

dora/src/masm/x64.rs
```diff
@@ -1119,15 +1137,15 @@ impl MacroAssembler {
 }
 
 impl MachineMode {
-    fn w(self) -> u8 {
+    pub fn w(self) -> u8 {
```
dinfuehr: Couldn't we get rid of this method and call `is64` instead?

soc:
@dinfuehr Added a second commit that migrates `w()` to `is64()`.

dinfuehr:
Thanks! Let's get rid of `w()` first and then do the refactoring in a separate PR.

soc:
@dinfuehr You mean swapping the two commits?

soc:
@dinfuehr Done!

dinfuehr:
> @dinfuehr You mean swapping the two commits?

Sorry I thought `w()` was introduced in this PR, that's why I wanted you to remove it in the first commit already. I think removing `w()` warrants it own PR, quite a lot of code changed and this shouldn't stop the new shift instructions from landing. Could you reduce this PR to the commit that introduces the shift instruction (but still uses `w()`)? Should just be a `push --force` of the previous commit.

soc:
Oh, ok – I actually thought this way around was cleaner, as it wasn't introducing new emits and then changing the signature again. Maybe I could create a PR just for the `w()` change and after that one is merged, create a new PR for the shift ops?

soc:
@dinfuehr I flipped the commits again!

dinfuehr:
Thanks! Could you move the second commit into another PR please?

soc:
@dinfuehr Done, will create 2nd PR after merge.

PR #145, opened 2020-01-29
Move "mode to w-bit" pattern match to function

soc:
As mentioned in https://github.com/dinfuehr/dora/pull/144.


dinfuehr:
Thanks! Not sure whether we still need this after the refactoring though.

soc:
Thanks!

PR #146, opened 2020-01-29
Fail when we are trying to install an intrinsic for a method not marked @internal

soc:
A few minutes too late for https://github.com/dinfuehr/dora/pull/131. :-)


dinfuehr:
Nice, thanks!

PR #147, opened 2020-01-31
Update to nightly-2020-01-08 and make use of the new Cargo.lock file format

soc:
See https://blog.rust-lang.org/2020/01/30/Rust-1.41.0.html#less-conflict-prone-cargolock-format.


dinfuehr:
Thanks!

PR #150, opened 2020-02-01
Test: Try to debug spurious CI failure in fannkuchredux1.dora

dinfuehr:


dinfuehr:
I've seen a spurious CI failure (happened twice) in tests/fannkuchredux1.dora. Can't reproduce it locally or on the CI when I run the tests in a loop ([output](https://github.com/dinfuehr/dora/runs/420563183?check_suite_focus=true)). IIRC it only happened when the fannkuchredux1.dora test ran first. Unfortunately the exit reason is missing, probably because the test was stopped with a signal. The next time this happens the reason should be in the log [now](https://github.com/dinfuehr/dora/commit/2b488900596115ede1f54c3eac47ccd7eee77f0b). If you see this failure please ping me.

PR #151, opened 2020-02-01
cannon: direct invoke

hzahradnik:
Until now, only DirectInvokeVoid was supported. I added the necessary code inside the bytecode generator and implemented the calls in cannon. Additionally, the call checks now for nil before calling the method.
Also I fixed a bug, where following code failed, because the object was not moved to the variable register and added a test case for it.
`let obj = Foo(1); // class Foo(let a: Int)
assert(obj.a == 1);
`

### DIFF
--- a/dora/src/bytecode/generator.rs
+++ b/dora/src/bytecode/generator.rs
@@ -1211,14 +1211,6 @@ impl DataDest {
         }
     }
 
-    fn is_reg(&self) -> bool {
-        match self {
-            DataDest::Effect => false,
-            DataDest::Reg(_) => true,
-            DataDest::Alloc => false,
-        }
-    }
-
     fn reg(&self) -> Register {
         match self {
             DataDest::Effect | DataDest::Alloc => panic!("not a register"),
--- a/dora/src/bytecode/generator.rs
+++ b/dora/src/bytecode/generator.rs
@@ -529,12 +529,14 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
         }
 
         if call_type.is_ctor_new() || call_type.is_ctor() {
-            if dest.is_reg() {
-                let return_reg = self.ensure_register(dest, BytecodeType::Ptr);
-                self.gen.emit_mov_ptr(return_reg, start_reg);
-                return_reg
-            } else {
-                start_reg
+            match dest {
+                DataDest::Effect => Register::invalid(),
+                DataDest::Reg(_) => {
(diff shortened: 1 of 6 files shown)


dora/src/cannon/codegen.rs
```diff
@@ -1101,9 +1101,47 @@ where
...
+        let position = self.bytecode.offset_position(self.current_offset.to_u32());
+        assert_eq!(bytecode_type_self, BytecodeType::Ptr);
+
+        self.asm.load_mem(
+            bytecode_type_self.mode(),
+            REG_RESULT.into(),
+            Mem::Local(offset_self),
+        );
+        self.asm
+            .test_if_nil_bailout(position, REG_RESULT.into(), Trap::NIL);
```
dinfuehr: AFAICS there is no test for the nil-check. Might be worth to add one.

dora/src/bytecode/generator.rs
```diff
@@ -1170,6 +1212,14 @@ impl DataDest {
         }
     }
 
+    fn is_reg(&self) -> bool {
```
dinfuehr: This isn't used anymore, right?

dora/src/bytecode/generator.rs
```diff
@@ -494,8 +528,16 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
             CallType::Intrinsic(_) => unreachable!(),
         }
 
-        if call_type.is_ctor_new() {
-            start_reg
+        if call_type.is_ctor_new() || call_type.is_ctor() {
+            match dest {
+                DataDest::Effect => Register::invalid(),
+                DataDest::Reg(_) => {
+                    let return_reg = self.ensure_register(dest, BytecodeType::Ptr);
```
dinfuehr: Is it necessary to invoke `ensure_register` here? I suppose DataDest::Reg(XXX) already has the right register.

dora/src/bytecode/generator.rs
hzahradnik: Currently not. Will remove it.

dora/src/bytecode/generator.rs
hzahradnik: You're right.

dora/src/cannon/codegen.rs
hzahradnik: As I plan to activate some existing tests in the near future, I could activate `nil3.dora`. This should be sufficient.

dinfuehr:
Cool, thanks!

PR #152, opened 2020-02-01
Drop nil from Vec, BitSet, String, Exception and Error

soc:
Additional pieces extracted from https://github.com/dinfuehr/dora/pull/128.

### DIFF
--- a/dora/src/object.rs
+++ b/dora/src/object.rs
@@ -13,6 +13,7 @@ use crate::gc::Address;
 use crate::handle::{root, Handle};
 use crate::mem;
 use crate::size::InstanceSize;
+use crate::stdlib::throw_native;
 use crate::vm::{ClassDefId, VM};
 use crate::vtable::VTable;
 
@@ -497,7 +498,8 @@ impl Str {
         let total_len = val.len();
 
         if offset > total_len {
-            return str_alloc_heap(vm, 0);
+            throw_native();
+            return Ref::null();
         }
 
         let len = std::cmp::min(total_len - offset, len);
@@ -521,7 +523,8 @@ impl Str {
 
             handle
         } else {
-            str_alloc_heap(vm, 0)
+            throw_native();
+            Ref::null()
         }
     }
 
--- a/dora/src/object.rs
+++ b/dora/src/object.rs
@@ -459,6 +459,10 @@ impl Str {
             + self.len() // string content
     }
 
+    pub fn empty(vm: &VM) -> Ref<Str> {
+        str_alloc_heap(vm, 0)
+    }
+
     /// allocates string from buffer in permanent space
     pub fn from_buffer_in_perm(vm: &VM, buf: &[u8]) -> Ref<Str> {
         let mut handle = str_alloc_perm(vm, buf.len());
@@ -493,7 +497,7 @@ impl Str {
         let total_len = val.len();
 
         if offset > total_len {
-            return Ref::null();
+            return str_alloc_heap(vm, 0);
         }
 
         let len = std::cmp::min(total_len - offset, len);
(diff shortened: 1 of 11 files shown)


soc:
After this PR there are no nils in the standard library left with the exception of https://github.com/dinfuehr/dora/pull/128/files#diff-fdd12e40f618e7564137f156b2f8ba69, which fails due to some bug in initialization.

dora/src/object.rs
```diff
@@ -517,7 +521,7 @@ impl Str {
 
             handle
         } else {
-            Ref::null()
```
dinfuehr: I don't think we should silently swallow this, better to throw an exception here or make the return type `Option[String]`.

dora/src/object.rs
```diff
@@ -493,7 +497,7 @@ impl Str {
         let total_len = val.len();
 
         if offset > total_len {
-            return Ref::null();
```
dinfuehr: Here as well

dinfuehr:
I think we should keep nil in those two places in this PR and fix it in a subsequent one.

dinfuehr:
Wondering about the initialization bug: Do you have a repro?

dora/src/object.rs
soc: This is handled in https://github.com/dinfuehr/dora/pull/152/files#diff-edf3ee5de6272371b58c0a2d02e7c112.

dora/src/object.rs
soc: This too: https://github.com/dinfuehr/dora/pull/152/files#diff-edf3ee5de6272371b58c0a2d02e7c112.

dora/src/object.rs
dinfuehr: Right sorry, haven't seen that... Couldn't we throw the exception directly from the native code instead? Then we wouldn't need the wrappers (fromStringPart and fromBytesPart) anymore and fromStringPartOrNull and fromBytesPartOrNull can simply replace them.

dinfuehr:
I am also wondering whether we should add special "magic" to the `Option` type in the runtime/compiler, such that `Option[SomeClass]` is already just the reference without the indirection and performance overhead. This also would make it easier to use Option types in native Rust code. We could remove part of this magic later when we get ADTs later.

soc:
> I am also wondering whether we should add special "magic" to the Option type in the runtime/compiler, such that Option[SomeClass] is already just the reference without the indirection and performance overhead. This also would make it easier to use Option types in native Rust code. We could remove part of this magic later when we get ADTs later.

I absolutely agree with this (though I'm not qualified to implement it). :-)

dora/src/object.rs
soc: Ah, sorry, I just migrated to native excpetions + options before reading this ... I'll push and then you can decide if it's ok or needs to be reworked.

soc: Does this look ok? https://github.com/dinfuehr/dora/pull/152/commits/77e8b3c5fd69f4f9fbdf87565d48b1dc65b0ddae

dinfuehr: Sure, even better ;)

dinfuehr:
> > I am also wondering whether we should add special "magic" to the Option type in the runtime/compiler, such that Option[SomeClass] is already just the reference without the indirection and performance overhead. This also would make it easier to use Option types in native Rust code. We could remove part of this magic later when we get ADTs later.
> 
> I absolutely agree with this (though I'm not qualified to implement it). :-)

Me too, I have no idea how hard this is going to be. I'll give it a try.

Very cool, thanks!

dora/src/object.rs
soc: Ah, great! Thanks!

PR #155, opened 2020-02-02
bytecode: fix method call

hzahradnik:
As you mentioned to me, the bytecode generator didn't handle method calls correctly. Instead of moving the called object to the corresponding argument register, it pushed `self` of the current function.

As currently `var_self` hasn't any check to assure the function has self, it doesn't throw an error, but returns the first variable (=`Register(0)`). Because of this, some tests coincidentally worked, as the register for the called object was `Register(0)`. I didn't changed this behavior for now, but added a test to assure the bytecode generator handles this correctly.

### DIFF
--- a/dora/src/bytecode/generator.rs
+++ b/dora/src/bytecode/generator.rs
@@ -424,18 +424,18 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
             Register::zero()
         };
         self.gen.set_position(expr.pos);
-        let arg_start_reg = if let CallType::CtorNew(ty, _) = &*call_type {
-            let cls_id = specialize_class_ty(self.vm, *ty);
-            self.gen.emit_new_object(start_reg, cls_id);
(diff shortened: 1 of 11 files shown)


dora/src/bytecode/generator.rs
```diff
@@ -424,18 +424,18 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
-            let self_reg = self.var_reg(self_id);
-
-            self.gen.emit_mov_ptr(start_reg, self_reg);
-            start_reg.offset(1)
-        } else {
-            start_reg
+        let arg_start_reg = match *call_type {
+            CallType::CtorNew(ty, _) => {
+                let cls_id = specialize_class_ty(self.vm, ty);
+                self.gen.emit_new_object(start_reg, cls_id);
```
dinfuehr: That set_position should probably move right before `emit_new_object`. You can do this now or in a subsequent PR.

PR #163, opened 2020-02-08
Various library work

soc:
(see individual commits)

### DIFF
--- a/dora/stdlib/Iterator.dora
+++ b/dora/stdlib/Iterator.dora
@@ -1,0 +1,4 @@
+trait Iterator {
+  fun hasNext() -> Bool;
+  fun next() -> Int; // this should be made generic
+}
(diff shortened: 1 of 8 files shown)


dora/stdlib/Iterator.dora
```diff
@@ -0,0 +1,4 @@
+trait Iterator {
+  fun hasNext() -> Bool;
+  fun next() -> Self;
```
soc: I think with this return type the impls for `IntRangeIter` and `CodepointIterator` should not compile, but they do.

I guess we are missing a check there, @dinfuehr?

dinfuehr: This is indeed surprising, most likely a bug.

soc: Shall I remove the commit from this PR for now?

dinfuehr: Yes.

soc: Ah, maybe better: I kept the file move and just undid the CodepointIterator impl and the result type change. Also acceptable?

soc: Repro for bug is https://github.com/dinfuehr/dora/issues/164

PR #165, opened 2020-02-10
bytecode/cannon: add intrinsics for binary enum intrinsics (==, !=) 

hzahradnik:


dora/src/bytecode/generator.rs
```diff
@@ -932,6 +934,11 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
                 BinOp::Cmp(CmpOp::Ge) => self.gen.emit_test_ge_char(dest, lhs_reg, rhs_reg),
                 _ => unreachable!(),
             },
+            Intrinsic::EnumEq | Intrinsic::EnumNe => match op {
```
hzahradnik: Is there a reason that there is an explicit `Intrinsic::EnumNe`? This seems to be redundant, as we also have the operand.

dinfuehr: I think you need to respect the intrinsic here, not the operand.

hzahradnik: Changed.

Issue #56, opened 2019-04-06
"Direct" implementation of traits


soc:
I think it would make sense to allow classes (and structs) to directly implement traits, such that

```
class Foo(...) {
  fun compareTo(rhs: Foo) -> Int = ...
}

impl Comparable for Foo {
  fun compareTo(other: Foo) -> Int = self.compareTo(other);
}
```

can be written as

```
class Foo(...) <keyword> Comparable {
  fun compareTo(rhs: Foo) -> Int = ...
}
```

In addition to that, I think it would make sense that `impl`s – whose fun-to-be-implemented exists with the exact signature in the `class` – do not require writing down the function if the classes' implementation matches:

```
class Foo(...) {
  fun compareTo(rhs: Foo) -> Int = ...
}

impl Comparable for Foo {} // no definition of compareTo necessary.
```

soc:
I think the hardest part is deciding how the class syntax should look like. I propose using either `with` or `supports` as the keyword, but I don't have a too strong opinion on that.

One problem is that if the class already has an inheritance relationship, the syntax would look kinda weird:

```
class Foo(...) : Bar(...) supports Comparable {
  fun compareTo(rhs: Foo) -> Int = ...
}
```

It feels like this creates a much stronger visual connection between `Bar` and `Comparable` than between `Foo` and `Comparable`, which I believe is undesirable.

On the other hand, I kind of dislike the use of `:` for different things, so maybe this is a good way to tackle this issue:

```
class Foo(...) extends Bar(...) supports Comparable {
  fun compareTo(rhs: Foo) -> Int = ...
}
```

Based on the exact keyword chosen for direct trait implementations, the naming scheme of traits would need to be adapted:

```
... supports Comparable
... supports Sortable
... supports Equals
... supports Identity

... with Comparable
... with Sortable
... with Equals
... with Identity
```

In both cases there are combinations that sound off.

Maybe there is a consistent selection of names, but I have a hard time figuring it out:
```
supports: Comparing, Sorting, Equality, Identity
with: Comparable, Sortable, Equatable, Identifiable
```

soc:
I have worked on this and added the required logic to the parser and ast parts.

I think the next step would be to add the direct implementations of traits to the VM's list of `impl`s.

@dinfuehr Do you have a suggestion at which point this should happen?

dinfuehr:
[...]

PR #166, opened 2020-02-11
Allow defining and calling methods without parentheses

soc:
### Why do this?

The idea is to allow decoupling how a member is implemented from how it is exposed,
allowing authors to reevaluate the costs of (re-)computation vs. storage
without breaking every existing caller of such code.

This approach is based on experience with earlier attempts, like

- Java's "use getters/setters everywhere, just in case"
- C#'s "use properties, which are source incompatible with getters/setters and binary incompatible with fields"

but provides the desired decoupling while _not_ breaking user code and avoiding other drawbacks of older designs as well.

As the distinction between fun, let and var is nevertheless important,
the idea is that an IDE should use different colors when referring to such members
(which provides more useful information than in Java's or C#'s case).

### When to use this?

The idea is that methods which are "constant" for the instance they are called on
are defined without (), and methods that mutate their instance are defined with ().

Example:
    `true.not, 1.toString, array.length, iterator.hasNext`
        vs.
    `iterator.next(), random.nextInt(), ...`

### Further steps

This is one of the steppingstones to allow classes to directly implement traits with their constructor parameters (https://github.com/dinfuehr/dora/issues/56):

```
trait HasAge {
  fun age -> Int;
}
class Person(let name: String, let age: Int) impl HasAge {
  // no further work to be done to implement HasAge
  // let age acts as an implementation for the trait's fun age
}
```

### DIFF
--- a/dora/src/bytecode/generator.rs
+++ b/dora/src/bytecode/generator.rs
@@ -1,8 +1,8 @@
-use dora_parser::lexer::position::Position;
 use std::collections::HashMap;
 use std::convert::TryInto;
 
 use dora_parser::ast::*;
+use dora_parser::lexer::position::Position;
 
 use crate::bytecode::{
     BytecodeBuilder, BytecodeFunction, BytecodeType, ConstPoolIdx, Label, Register,
@@ -10,6 +10,7 @@ use crate::bytecode::{
 use crate::semck::specialize::specialize_type;
 use crate::semck::{expr_always_returns, expr_block_always_returns};
 use crate::ty::{find_impl, SourceType, SourceTypeArray};
+use crate::utils::iter_some;
 use crate::vm::{
     AnalysisData, CallType, ConstId, EnumId, Fct, FctId, GlobalId, IdentType, Intrinsic, StructId,
     TupleId, VarId, VM,
@@ -75,7 +76,7 @@ impl<'a> AstBytecodeGen<'a> {
             }
         }
 
-        for param in &ast.params {
+        for param in iter_some(&ast.params) {
             let var_id = *self.src.map_vars.get(param.id).unwrap();
             let ty = self.var_ty(var_id);
 
@@ -932,6 +933,25 @@ impl<'a> AstBytecodeGen<'a> {
         if let Some(struct_id) = object_ty.struct_id() {
             let type_params = object_ty.type_params(self.vm);
             return self.visit_expr_dot_struct(expr, struct_id, type_params, dest);
+        }
+
+        let call_type = self.src.map_calls.get(expr.id);
+        if call_type.map(|c| c.is_method()).contains(&true) {
+            let dot = ExprDotType {
+                id: expr.id,
+                pos: expr.pos,
+                span: expr.span,
+                lhs: expr.lhs.clone(),
+                rhs: expr.rhs.clone(),
+            };
+            let call = ExprCallType {
+                id: expr.id,
+                pos: expr.pos,
+                span: expr.span,
+                callee: Box::new(Expr::Dot(dot)),
+                args: Vec::new(),
+            };
+            return self.visit_expr_call(&call, dest);
         }
 
         let (cls_ty, field_id) = {
--- a/dora-parser/src/ast.rs
+++ b/dora-parser/src/ast.rs
@@ -4,7 +4,6 @@ use std::sync::Arc;
 
 #[cfg(test)]
 use crate::interner::Interner;
-
 use crate::interner::Name;
 use crate::lexer::position::{Position, Span};
 use crate::lexer::token::{FloatSuffix, IntBase, IntSuffix};
@@ -667,7 +666,7 @@ pub struct Function {
     pub internal: bool,
     pub is_constructor: bool,
 
-    pub params: Vec<Param>,
+    pub params: Option<Vec<Param>>,
 
     pub return_type: Option<Type>,
     pub block: Option<Box<ExprBlockType>>,
--- a/dora-parser/src/ast/dump.rs
+++ b/dora-parser/src/ast/dump.rs
@@ -1,7 +1,6 @@
 use std::sync::Arc;
 
 use crate::ast::*;
-
 use crate::interner::{ArcStr, Interner, Name};
 
 macro_rules! dump {
@@ -372,12 +371,15 @@ impl<'a> AstDumper<'a> {
             dump!(d, "internal = {}", fct.internal);
             dump!(d, "params");
             d.indent(|d| {
-                if fct.params.is_empty() {
-                    dump!(d, "no params");
-                } else {
-                    for param in &fct.params {
-                        d.dump_param(param);
+                if fct.params.is_some() {
+                    let params = fct.params.as_ref().unwrap();
+                    if !params.is_empty() {
+                        for param in params {
+                            d.dump_param(param);
+                        }
                     }
+                } else {
+                    dump!(d, "no params");
                 }
             });
 
--- a/dora-parser/src/ast/visit.rs
+++ b/dora-parser/src/ast/visit.rs
@@ -204,7 +204,7 @@ pub fn walk_field<V: Visitor>(v: &mut V, f: &Field) {
 }
 
 pub fn walk_fct<V: Visitor>(v: &mut V, f: &Function) {
-    for p in &f.params {
+    for p in iter_some(&f.params) {
         v.visit_param(p);
     }
 
@@ -381,5 +381,12 @@ pub fn walk_expr<V: Visitor>(v: &mut V, e: &Expr) {
         Expr::LitStr(_) => {}
         Expr::LitBool(_) => {}
         Expr::Ident(_) => {}
+    }
+}
+
+fn iter_some<T>(opt: &Option<Vec<T>>) -> impl Iterator<Item = &T> {
+    match opt {
+        Some(opt) => opt.iter(),
+        None => (&[]).iter(),
     }
 }
--- a/dora-parser/src/builder.rs
+++ b/dora-parser/src/builder.rs
@@ -1,8 +1,6 @@
 use crate::ast::*;
-
 use crate::interner::*;
 use crate::lexer::position::{Position, Span};
-
 use crate::parser::NodeIdGenerator;
 
 pub struct Builder<'a> {
@@ -81,7 +79,7 @@ pub struct BuilderFct<'a> {
     is_public: bool,
     is_constructor: bool,
     return_type: Option<Type>,
-    params: Vec<Param>,
+    params: Option<Vec<Param>>,
     block: Option<Box<ExprBlockType>>,
 }
 
@@ -94,7 +92,7 @@ impl<'a> BuilderFct<'a> {
             is_public: false,
             is_constructor: false,
             return_type: None,
-            params: Vec::new(),
+            params: None,
             block: None,
         }
     }
@@ -108,9 +106,12 @@ impl<'a> BuilderFct<'a> {
     ) -> &mut BuilderFct<'a> {
         let id = self.id_generator.next();
 
+        if self.params.is_none() {
+            self.params = Some(Vec::new());
+        }
         let param = Param {
             id,
-            idx: self.params.len() as u32,
+            idx: self.params.as_ref().unwrap().len() as u32,
             name,
             variadic,
             pos,
@@ -118,7 +119,7 @@ impl<'a> BuilderFct<'a> {
             data_type: ty,
         };
 
-        self.params.push(param);
+        self.params.as_mut().unwrap().push(param);
         self
     }
 
--- a/dora-parser/src/parser.rs
+++ b/dora-parser/src/parser.rs
@@ -6,9 +6,7 @@ use crate::ast;
 use crate::ast::*;
 use crate::builder::Builder;
 use crate::error::{ParseError, ParseErrorAndPos};
-
 use crate::interner::*;
-
 use crate::lexer::position::{Position, Span};
 use crate::lexer::reader::Reader;
 use crate::lexer::token::*;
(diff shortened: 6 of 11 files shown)


dora/src/baseline/codegen.rs
```diff
@@ -1166,19 +1167,18 @@ where
...
+        match ident_type {
+            &IdentType::Field(ty, field) => {
+                let ty = self.specialize_type(ty);
 
-        self.emit_expr(&expr.lhs, REG_RESULT.into());
-        self.emit_field_access(expr.pos, ty, field, REG_RESULT, dest.any_reg());
+                self.emit_expr(&expr.lhs, REG_RESULT.into());
+                self.emit_field_access(expr.pos, ty, field, REG_RESULT, dest.any_reg());
+            }
+            &IdentType::Method(ty, method) => unimplemented!(),
```
soc: This is where we currently fail – I'm unsure we should even end up with an `ExprDotType` and in `emit_dot` in the first place.

soc:
@dinfuehr What do you think?

dinfuehr:
> The goal is to allow defining and calling methods without the ().
> The idea is that methods which are "constant" for the instance
> they are called on are defined without (),
> and methods that mutate their instance use ().

It is unfortunate that this can only be a convention in Dora but not something we could enforce in the compiler..

> (And allowing people to switch the implementation easily, by changing the keyword, not by also making them add/remove the `()` etc.)

This could be done by the IDE just like the coloring you mention later.

> * C#'s "use properties, which are source incompatible with getters/setters and binary incompatible with fields"

How does this solution solve binary compatibility? Right now it looks like you have to recompile as well.

And what would you do about setters? Right now I only see getters handled and AFAICT you didn't mention them yet.

About the implementation: I don't think the AST and parser need to change at all here. typeck/expr.rs needs to recognize <expr>.<.methodname> as function call even though there are no parens `()`. Then baseline/codegen.rs can simply pick up this information and generate the call.

soc:
> It is unfortunate that this can only be a convention in Dora but not something we could enforce in the compiler..

Agreed! In my experience the convention has worked reasonably well elsewhere, so I'm hoping it is working this way in Dora too.

> This could be done by the IDE just like the coloring you mention later.

Sounds good!

> How does this solution solve binary compatibility? Right now it looks like you have to recompile as well.

The idea is that we get rid of this distinction in non-source formats too, that Dora might get. Such that the meaning of `foo.bar` is not hard-coded into such a format, but something the runtime deals with during linking such formats.

> And what would you do about setters? Right now I only see getters handled and AFAICT you didn't mention them yet.

There is a possibility to add a rule for setters too, like `foo.setBar(x)` ~= `foo.bar = x` (excluding "the implement `fun` with `var`" part), similar to how we deal with `.get` and `.set`, but as it has a higher language surface than this, I'd first try to collect some experience on whether it would be worth it.

> About the implementation: I don't think the AST and parser need to change at all here. typeck/expr.rs needs to recognize .<.methodname> as function call even though there are no parens `()`. Then baseline/codegen.rs can simply pick up this information and generate the call.

Thanks! I will try it that way!

dinfuehr:
> The idea is that we get rid of this distinction in non-source formats too, that Dora might get. Such that the meaning of `foo.bar` is not hard-coded into such a format, but something the runtime deals with during linking such formats.

Dora kinda already has this non-source format with Cannon and the Bytecode. This should be pretty similar to how the binary format would look like. I guess you would have to assume that all those field accesses are actually method calls.

> There is a possibility to add a rule for setters too, like `foo.setBar(x)` ~= `foo.bar = x` (excluding "the implement `fun` with `var`" part), similar to how we deal with `.get` and `.set`, but as it has a higher language surface than this, I'd first try to collect some experience on whether it would be worth it.

Not really a fan of this setter magic. I mean we do something similar for operator overloading but this still feels wrong. Does it even make sense to even do the getter thing as long as the setter story isn't better? What happens if you have a both `bar` and `setBar` in the same class? I guess that needs to be prevented. Again, every statement `obj.field = value;` has to be compiled as `obj.setField(value)` - just in case someone might replace it later.

I would assume that only a small subset of classes are actually public and used outside of a library. If you ask me its the authors obligation to use the appropriate language feature (field, method, property, etc.) in those cases to allow for changes without breaking user-code. Even if one could solve this issue perfectly in the language there are numerous other subtle ways to break code that can't be solved in the language (e.g. would we need to have the same thing for global variables and functions as well). IMHO we shouldn't base language design decisions on this.

soc:
> I guess you would have to assume that all those field accesses are actually method calls.

Yes. I'd try to delay the actual decision such that the determination whether `.foo` is a method or a field is not reflected in any kind of publicly distributed bytecode.

> Not really a fan of this setter magic. I mean we do something similar for operator overloading but this still feels wrong. Does it even make sense to even do the getter thing as long as the setter story isn't better?

I'm also not a fan of the setter magic (I just mentioned it for completeness) – I'd propose not doing it at all for setters.

> If you ask me its the authors obligation to use the appropriate language feature (field, method, property, etc.) in those cases to allow for changes without breaking user-code.

Isn't that what Java tried – with the result that everything has a getter in front of it and no way to tell anymore whether `getFoo` is a simple getter for a field or a hugely expensive computation.

> Even if one could solve this issue perfectly in the language there are numerous other subtle ways to break code that can't be solved in the language (e.g. would we need to have the same thing for global variables and functions as well).

I'd say that the approach is sound in general – the same rule for instances would apply to modules (which are intended to get rid of global variables and functions). If there are addiitonal ways that could pose issues, I'd love to work on them.

dinfuehr:
> Isn't that what Java tried – with the result that everything has a getter in front of it and no way to tell anymore whether `getFoo` is a simple getter for a field or a hugely expensive computation.

Yes, I would say that this was certainly considered good practice and the "Java style" (fwiw I am guilty myself of doing this as well 😆). I would say the difference to Java back then is that in Java this was encouraged for each and every class (including internal classes). Languages like Rust also have this difference between fields and methods and get along just fine. In Rust you might want to use getters/setters instead of fields in public classes when you want the option of changing the implementation later without breaking user-code. There might be other cases as well, but it's definitely a lot less frequent than in Java.
 
> I'd say that the approach is sound in general – the same rule for instances would apply to modules (which are intended to get rid of global variables and functions). If there are addiitonal ways that could pose issues, I'd love to work on them.

One example that comes to mind: In Rust usually it is a compile error if you don't match all variants of an enum - which I consider pretty useful. If the enum is now defined in a library you can't add a variant without breaking user-code. That's why they've recently introduced an attribute for such use cases. (The goal for Dora is also to get support for ADTs)

Since you also strive for ABI-compatibility: This becomes much harder with value types. We now have those in Dora as well with tuples. With classes it is easy to add an internal field - this is also fine for the ABI/bytecode. Simply because we only pass the references to objects and the internal field is hidden away in the constructor or methods. Now imagine someone wanting to save the memory allocation and use a tuple as argument. Adding a field later to a tuple is much harder - especially if you want to keep the ABI compatible. We would need something like Swift does for ABI-compatibility and this might just be too much work for us. That problem only becomes worse when we also implement structs and ADTs as value types (not sure anymore if the latter is possible in Dora with concurrent marking). This means that an API that is supposed to be extensible, should probably favor classes over tuples.

For more examples I would need to dig deeper. My main point still stands: Designing good and extensible APIs is hard and takes care. We cannot solve all (or even most) of these issues in the language. ABI-compatibility is a whole different matter (not only do we have value types but also generics!?) and we can't solve semantic differences that might break user-code as well (e.g. one of the reasons inheritance is not recommended for code reuse). IMHO this proposal also only solves part of this particular problem by ignoring setters. It also comes with the drawback that you can't figure out anymore from `obj.x` whether this is a potentially costly method call (even if it doesn't change the current class) or a simple field (you criticised this above for Java). Personally I am not really convinced that what this features really solves outweighs this disadvantage.

soc:
> Languages like Rust also have this difference between fields and methods and get along just fine. In Rust you might want to use getters/setters instead of fields in public classes when you want the option of changing the implementation later without breaking user-code. There might be other cases as well, but it's definitely a lot less frequent than in Java.

I believe that this is the right choice for languages like Rust, C, C++ that were designed around the paradigm of cost-free abstraction, but what's almost a requirement for such languages does necessarily translate to languages with runtimes and GC like Dora.

> One example that comes to mind: In Rust usually it is a compile error if you don't match all variants of an enum - which I consider pretty useful.

I agree on this – but how does this relate to making `()` optional for functions?

I also agree on your later paragraphs on why binary compatibility is hard – but I feel this does not detract from this proposal, which focuses on solving a well-understood issue in a way that is demonstrably better than other attempts in the past.

> IMHO this proposal also only solves part of this particular problem by ignoring setters.

The proposal is about `fun`s _without parameters_ – it doesn't apply to setters, since they have parameters, just as it doesn't apply to functions with 2, 3 or 10 parameters, because none of them are replaceable by a `let`.

> It also comes with the drawback that you can't figure out anymore from obj.x whether this is a potentially costly method call (even if it doesn't change the current class) or a simple field (you criticised this above for Java).

Now I'm confused – isn't it the complete opposite? One of the core benefits of the proposal is that it allows to see whether `obj.x` is a field access or a method call _without_ forcing us to sacrifice source or binary compatibility for it, like done in many other languages.

---

I didn't intend this to be a controversial proposal – it felt like a cheap, easy win with no drawbacks for me. Maybe I have explained things poorly, then I'll focus on doing better.

soc:
@dinfuehr I got things working – except that the compiler isn't as strict as it should be:
It currently accepts `false.toString()` even if the method is defined as `fun toString` (the other way around has the same issue).

This is probably due to not retaining the "with ()" vs. "without ()" in `Fct` – I htink you mentioned earlier that it wasn't necessary to have this info in `Fct` – do you have a different way od dealing with this?

soc:
The last commit that implements stricter checking of () fails with

```
thread 'main' panicked at 'slice index starts at 1 but ends at 0', src/libcore/slice/mod.rs:2680:5
stack backtrace:
   0:     0x5623d8e3d134 - backtrace::backtrace::libunwind::trace::ha0e5d4d0df9559a5
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1:     0x5623d8e3d134 - backtrace::backtrace::trace_unsynchronized::h86d5818cd0848dcd
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2:     0x5623d8e3d134 - std::sys_common::backtrace::_print_fmt::h546442ce2690d8f8
                               at src/libstd/sys_common/backtrace.rs:77
   3:     0x5623d8e3d134 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hc5c7f2f3dc7f3a8f
                               at src/libstd/sys_common/backtrace.rs:59
   4:     0x5623d8e5c14c - core::fmt::write::h1b228a8d0839cf13
                               at src/libcore/fmt/mod.rs:1052
   5:     0x5623d8e3a4e7 - std::io::Write::write_fmt::h18ba737c2eb4cde5
                               at src/libstd/io/mod.rs:1428
   6:     0x5623d8e3f325 - std::sys_common::backtrace::_print::hc0def0d4510194bf
                               at src/libstd/sys_common/backtrace.rs:62
   7:     0x5623d8e3f325 - std::sys_common::backtrace::print::hdbda01cface40d57
                               at src/libstd/sys_common/backtrace.rs:49
   8:     0x5623d8e3f325 - std::panicking::default_hook::{{closure}}::h3340d810a97d665d
                               at src/libstd/panicking.rs:204
   9:     0x5623d8e3f011 - std::panicking::default_hook::h7bf6e61beb8aa6f0
                               at src/libstd/panicking.rs:224
  10:     0x5623d8e3f98a - std::panicking::rust_panic_with_hook::h4a97e0c9dc5a4ffb
                               at src/libstd/panicking.rs:472
  11:     0x5623d8e3f570 - rust_begin_unwind
                               at src/libstd/panicking.rs:380
  12:     0x5623d8e5a2e1 - core::panicking::panic_fmt::h3fb74ee9e4994c2e
                               at src/libcore/panicking.rs:85
  13:     0x5623d8e5d8c6 - core::slice::slice_index_order_fail::h67387fcf7d5126ac
                               at src/libcore/slice/mod.rs:2680
  14:     0x5623d8479838 - <core::ops::range::Range<usize> as core::slice::SliceIndex<[T]>>::index::h65b52f104c1e0117
                               at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/slice/mod.rs:2836
  15:     0x5623d8478370 - <core::ops::range::RangeFrom<usize> as core::slice::SliceIndex<[T]>>::index::h2a49157c1b0ce17b
                               at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/slice/mod.rs:2915
  16:     0x5623d8472e88 - core::slice::<impl core::ops::index::Index<I> for [T]>::index::hd025f466a6867253
                               at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/slice/mod.rs:2656
  17:     0x5623d83c05db - <alloc::vec::Vec<T> as core::ops::index::Index<I>>::index::he6c2ea27e292144f
                               at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/liballoc/vec.rs:1883
  18:     0x5623d86560ad - dora::vm::fct::Fct::params_without_self::h7c9f6d295dcf9d83
                               at dora/src/vm/fct.rs:189
  19:     0x5623d85ecd6f - dora::semck::superck::check_fct_modifier::he5eddee11e2201c3
                               at dora/src/semck/superck.rs:229
  20:     0x5623d85ebd14 - dora::semck::superck::check_override::h1e139ece601877e9
                               at dora/src/semck/superck.rs:166
  21:     0x5623d859ba16 - dora::semck::check::h4de6ad4e4253cb1f
                               at dora/src/semck.rs:97
  22:     0x5623d83987e8 - dora::driver::start::start::h128332435f668d5f
                               at dora/src/driver/start.rs:47
```

soc:
The implementation works now – here is a rough draft of updating the standard library accordingly: https://github.com/dinfuehr/dora/compare/master...soc:topic/no-method-parens-std?expand=1

soc:
@dinfuehr I dropped everything including the checking of definition-site vs. use-site of nullary vs. nilary methods.
What's included now is only the first commit, without any std library changes.
Could you have a look?

dora/src/bytecode/generator.rs
```diff
@@ -934,6 +935,25 @@ impl<'a> AstBytecodeGen<'a> {
...
+            let call = ExprCallType {
+                id: expr.id,
+                pos: expr.pos,
+                span: expr.span,
+                callee: Box::new(Expr::Dot(dot)),
+                args: Vec::new(),
+            };
+            return self.visit_expr_call(&call, dest);
+        }
+
```
soc, line 956: This is rather questionable, but way shorter than the code before that did everything manually.

soc:
I converted this into an issue: https://github.com/dinfuehr/dora/issues/284

PR #168, opened 2020-02-14
cannon: support invocation of virtual calls

hzahradnik:
As discussed I discontinued my effort to implement the various operations on types and focused to get some of the benchmarks to run. So I implemented the virtual method invocation, which was already defined as bytecode and as visitor functions, but the generator and cannon didn't implement it or not correctly.

The changes are not really complex, mostly some refactoring in the bytecode generator and the implementation details of the corresponding virtual invocation in cannon. At least I think that, maybe I missed the bigger picture.

### DIFF
--- a/dora/src/bytecode/generator.rs
+++ b/dora/src/bytecode/generator.rs
@@ -387,7 +387,7 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
         let fct = self.vm.fcts.idx(fct_id);
         let fct = fct.read();
 
-        let callee_id = if fct.kind.is_definition() {
+        let callee_id = if fct.kind.is_definition() && !fct.is_virtual() {
             unimplemented!()
         } else {
             fct_id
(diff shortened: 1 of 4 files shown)


dora/src/bytecode/generator.rs
```diff
@@ -542,6 +520,92 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
+                BytecodeType::Ptr => self
+                    .gen
+                    .emit_invoke_virtual_ptr(return_reg, callee_id, start_reg, num_args),
+            }
+        }
+    }
+
+    fn visit_call_direct(
+        &mut self,
+        return_type: BuiltinType,
```
dinfuehr: It might make sense to move `return_type` down to `return_reg`.

dora/src/bytecode/generator.rs
```diff
@@ -542,6 +520,92 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
         }
     }
 
+    fn visit_call_virtual(
+        &mut self,
+        return_type: BuiltinType,
```
dinfuehr: Here too

PR #169, opened 2020-02-16
Add HashSet implementation

soc:
Shamelessly stolen from the HashMap code.

### DIFF
--- a/dora/stdlib/HashSet.dora
+++ b/dora/stdlib/HashSet.dora
@@ -1,0 +1,167 @@
+class HashSet[K: Hash + Equals] {
+    // BitSet.size == capacity * 2
+    // [bit 0: inserted; bit 1: deleted] * capacity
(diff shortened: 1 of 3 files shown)


dinfuehr:
I planned to implement `HashSet[K]` as a simple wrapper around `HashMap[K, ()]` instead of duplicating this code. I haven't done this so far since I didn't expect `()` to work in all places yet.  I've recently worked a bit on tuples, so I think I now have a better understanding of how much work this entails: I assume it is not too hard and not too much work.

(I think we would e.g. need to skip `()` as argument on calls and not do anything on assignments. Basically add `is_unit()` conditionals where I recently introduced `is_tuple()`. `Array[()]` being the most tricky part.)

If you want to tackle this I could definitely guide you. We should definitely merge this as-is though (whether you want to work on it or not). Otherwise I would implement this myself soon.

PR #174, opened 2020-02-22
cannon: long intrinsics (+,-,*,/,%)

hzahradnik:


dinfuehr:
Let's wait with merging this until all tests are enabled. What do you think?

hzahradnik:
Now all long intrinsics should be supported. While working on this, I saw, that the call to `not()` in `tests\cannon\int1.dora` is not executed. I could add the fix in this PR, but if you want, I can create a new PR.

dinfuehr:
Whatever you prefer...

PR #175, opened 2020-02-22
bytecode/cannon: store globals

hzahradnik:


dora/src/cannon/codegen.rs
```diff
@@ -904,6 +904,31 @@ where
             .store_mem(bytecode_type.mode(), Mem::Local(offset), reg);
     }
 
+    fn emit_store_global_field(&mut self, src: Register, global_id: GlobalId) {
```
dinfuehr: Why `field`? Could you please rename this.

hzahradnik: I copied the `emit_load_global_field` method header and just renamed it. Renamed both methods and removed the `field` part of the name.

dinfuehr:
Thanks for renaming the load functions as well!

PR #176, opened 2020-02-22
Add library code for Zero trait

soc:
This is the library part for the `Zero` trait – there is a small bit of compiler support necessary right now that records during typechecking whether a class (+ generics) supports zero or not.

The difference between `Default` and `Zero` is that `Default` can be implemented by user code in whatever fashion desired – while `Zero` is more of a marker that any zero-initialized memory of the appropriate size constitutes a valid instance.

This pretty much means only `structs` who do not contain any reference types¹ can be `Zero` – which is good enough for numbers (which are all "honorary" structs, but declared as classes currently), `Option`¹, probably time-/date-related stuff and much more .

¹ We pretend we already live in a world without `nil` here.
² If one squints hard enough.

### DIFF
--- a/dora/stdlib/Zero.dora
+++ b/dora/stdlib/Zero.dora
@@ -1,0 +1,59 @@
+// Implementations of the `Zero` trait indicate that zero-initialized memory of the appropriate size can be allocated to
+// represent a valid value of that type.¹
+//
+//   let numbers: Array[Int] = Array.ofZero[Int](4)
+//   assert(numbers == Array(0, 0, 0, 0)
+//
+//   let strings: Array[Option[String]] = Array.ofZero[Option[String]](2)
+//   assert(strings == Array(None, None)
+//
+// `Zero` is a special trait whose implementations are automatically derived for qualifying types.
+// It cannot be implemented manually.
+// A type qualifies if it is either `Option`, or obeys all of the following rules ...
+// - The type is a struct.
+// - The type does not contain any reference-types directly or transitively.
+//
+// ¹ This means that implementations of `Zero` may make full use of the operating system's optimization of `calloc`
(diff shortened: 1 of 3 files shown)


dora/stdlib/Zero.dora
```diff
@@ -0,0 +1,59 @@
...
+// - The type does not contain any reference-types directly or transitively.
+//
+// ¹ This means that implementations of `Zero` may make full use of the operating system's optimization of `calloc`
+//   (lazy allocation, lazy zeroing, copy on write, etc.), unlike implementations of the `Default` trait –
+//   which invoke an user-defined function to initialize each individual element (malloc & loop with function calls).
+
+
+// @restricted / @internal
+trait Zero {
+  @static fun zero() -> Self; // should be `let` instead of `fun`
```
dinfuehr: Do we need the `zero` method? I don't think we need it.

dora/stdlib/Zero.dora
```diff
@@ -0,0 +1,59 @@
...
+}
+
+impl Zero for Bool {
```
dinfuehr: I think we should implement the trait for all possible types automatically in the runtime in the future (no impl for Zero in the source). For this PR I think we can keep it this way, since this already allows us to use `arrayZero` without additional work.

soc: > I think we should implement the trait for all possible types automatically in the runtime in the future (no impl for Zero in the source).

Yes, that was my plan!

> For this PR I think we can keep it this way, since this already allows us to use arrayZero without additional work.

Agreed.

dora/stdlib/Zero.dora
soc: Might be useful for some non-zero-initialization stuff, too, I think. But can leave it out and figure this out at a later point in time.

dinfuehr: Agreed, might be useful for initialization of values...

PR #178, opened 2020-02-26
Add runtime support for Zero trait

soc:
@dinfuehr Where should I add that check `if cls.is_zero_initializable {add_zero_trait()/add_zero_impl() }`?

(I could also patch the "well-known" classes like `Int`, `Long` etc. immediately in `prelude::internal_classes`, but I think that would scale poorly as soon as users can define their own zero-initializable structs.)

### DIFF
--- a/dora/src/semck/typeparamck.rs
+++ b/dora/src/semck/typeparamck.rs
@@ -118,7 +118,7 @@ impl<'a, 'ast> TypeParamCheck<'a, 'ast> {
         let cls = cls.read();
 
         for &trait_bound in &tp.trait_bounds {
-            if !cls.traits.contains(&trait_bound) {
+            if !cls.implements_trait(self.vm, trait_bound) {
                 self.fail_trait_bound(trait_bound, ty);
                 succeeded = false;
             }
--- a/dora/src/vm/class.rs
+++ b/dora/src/vm/class.rs
@@ -195,6 +195,10 @@ impl Class {
             }
         }
     }
+
+    pub fn implements_trait(&self, vm: &VM, trait_id: TraitId) -> bool {
+        self.traits.contains(&trait_id) || vm.vips.zero_trait == trait_id && !self.ty.is_cls()
+    }
 }
 
 pub fn find_field_in_class(
--- a/dora/src/semck.rs
+++ b/dora/src/semck.rs
@@ -257,7 +257,7 @@ pub fn read_type<'ast>(vm: &VM<'ast>, file: FileId, t: &'ast Type) -> Option<Bui
                                 let cls = cls.read();
 
                                 for &trait_bound in &tp.trait_bounds {
-                                    if !cls.traits.contains(&trait_bound) {
+                                    if !cls.implements_trait(vm, trait_bound) {
                                         let bound = vm.traits[trait_bound].read();
                                         let name = ty.name(vm);
                                         let trait_name = vm.interner.str(bound.name).to_string();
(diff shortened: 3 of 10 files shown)


dinfuehr:
Not sure yet what the best way to implement is... I thought that we could adapt `implements_trait` to have an explicit check for the zero trait. So there would be no explicit flag, since we should be able to figure this out from the type alone.

soc:
@dinfuehr Alright, will do! I was concerned about special-casing it in such a way, but I guess we won't have dozens of traits that are going to be special-cased, so I think this will be manageable. Thanks!

soc:
@dinfuehr `implements_trait` was never called in that code (but `type_against_definition` in typeparamck); I replaced the various calls by a shared method on `Class`.

dora/src/vm/class.rs
```diff
@@ -76,6 +76,10 @@ pub struct Class {
     // true if this class is the generic Array class
     pub is_array: bool,
     pub is_str: bool,
+    // temporary measure as classes are never zero-initializable – but as Int, Long, Float, etc. are
+    // currently "declared" as classes, not structs, so we use this field until it's possible to
+    // migrate them to structs.
+    pub is_zero_initializable: bool,
```
dinfuehr: I don't think we need this. `cls.ty` will give you the BuiltinType for a class, if this is `BuiltinType::Class(_)` then this is a *real* class. Int/Long/Float/etc. will have BuiltinType::Int/Long/Float/etc. instead.

dinfuehr: This would also mean that you don't need to change anything in prelude.rs (apart from initializing zero_trait_id).

soc: Ok, will change it.

dora/src/semck/typeparamck.rs
```diff
@@ -118,7 +118,7 @@ impl<'a, 'ast> TypeParamCheck<'a, 'ast> {
         let cls = cls.read();
 
         for &trait_bound in &tp.trait_bounds {
-            if !cls.traits.contains(&trait_bound) {
+            if !cls.implements_trait(self.vm, trait_bound) {
```
dinfuehr: This works for classes but not for tuples. Tuples never implement Zero, but we can also solve this in the next PR.

dora/src/vm/class.rs
soc: (THe idea was that this gave us some more flexibility for the case where e. g. `Option`'s `none` case was compiled down to zeroes, but the definition wasn't changed from class to struct...

dinfuehr:
I think you could also add a test to typeck/tests.rs where we ensure that classes do not implement zero and e.g. invoke `fun foo[T: Zero](a: T) {}` with a String which should result in an error.

dora/src/vm/class.rs
dinfuehr: Maybe we don't need the flag in those cases as well. We might be able to just check the memory layout of the ADT and don't need a separate flag.

soc:
@dinfuehr updated!

dinfuehr:
Very cool, thanks!

dinfuehr:
I realized that most likely the `zero()` method within the Zero trait isn't working right now. If we want to allow use cases like `fun f() -> Int { Int::zero() }` or `fun f[T: Zero]() { return T::zero(); }` we need to implement/test this as well. Alternatively we could also get rid of that method.. What do you think?

soc:
> I realized that most likely the zero() method within the Zero trait isn't working right now.

Agreed.

> What do you think?

I'd be happy to implement it, but I'm a bit unsure how to do it with the chosen approach; as we neither have classes implementing the trait (in the classic sense) nor any impls that could implement the method.

dora/src/vm/class.rs
soc: I assumed it would make sense if we didn't recompute this information every time we need it, but only computed it once per class and saved it there.

PR #179, opened 2020-02-26
bytecode: emit InvokeStatic for method calls without nil check (e.g. Int.toString())

hzahradnik:
Currently the bytecode generator didn't check if a call of type `CallType::Method` really needed the nil check, which is necessary for direct invocation. As a result `1.toString()` was translated to `InvokeDirectPtr(...)`. But when cannon tried to generate the machine code, it checked if the first argument ("self") was indeed of type `BytecodeType::Ptr`. This ultimately failed, as the first argument is `Int`.

I added a check in the generator for`CallType::Method` and if the first argument type is not a Ptr, a flag is set. If the flag is set, `InvokeStatic` is emitted.

### DIFF
--- a/dora/src/bytecode/generator.rs
+++ b/dora/src/bytecode/generator.rs
@@ -400,7 +400,7 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
         let fct = self.vm.fcts.idx(fct_id);
         let fct = fct.read();
 
-        if let FctKind::Builtin(_intrinsic) = fct.kind {
+        if (fct.kind.is_definition() && !fct.is_virtual()) || fct.kind.is_intrinsic() {
             unimplemented!()
         }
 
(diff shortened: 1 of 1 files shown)


dora/src/bytecode/generator.rs
```diff
@@ -437,18 +428,23 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
             Register::zero()
         };
         self.gen.set_position(expr.pos);
-        let arg_start_reg = match *call_type {
+        let (arg_start_reg, nil_check) = match *call_type {
```
dinfuehr: I think you can move that `nil_check` variable down into the next match for `CallType::Method`.

hzahradnik: Yep, will change it, when I come home.

dinfuehr:
Thanks!

PR #181, opened 2020-03-06
bytecode/cannon/vm: Add specializations for functions to vm and use in bytecode/cannon

hzahradnik:
As discussed, to support function and class type params when calling a method, I added a `FctDef`. This is similar to `ClassDef` the specialization of a `Fct` with the corresponding type params.

The bytecode now uses the `FctDefId` instead of `FctDef` when referencing to a function. Cannon then uses this id to retrieve the corresponding type params and will generate the machine code accordingly.


dinfuehr:
Thanks!

Issue #118, opened 2019-12-28
FloatCmp/DoubleCmp intrinsic not handled


soc:
```
(5.0).compareTo(7.0)
```

fails with

```
thread 'main' panicked at 'unknown intrinsic DoubleCmp',
  dora/src/baseline/codegen.rs:2050:18
```

because the intrinsic is missing there.

This issue should be tackled after #115 is merged to avoid unnecessary merge conflicts.

soc:
I guess an additional problem is that `sortsAs` uses total order such that we only have 3 possible returns (before/same/after), while `compareTo` has four (smaller/equal/greater/unordered), which maps rather poorly to integers.

We could of course do something like assigning unordered `Int::minValue()`, but whatever we pick, it feels it's dangerous as people might unintentionally do `let z = x.compareTo(y); if z < 0 { ... }` intending that the `true` case only means "smaller".

soc:
This was fixed in https://github.com/dinfuehr/dora/commit/41817ae9b0505d551a2aefa795db59b31d850532 by @dinfuehr.

Thanks!

PR #182, opened 2020-03-07
Implement {Float,Double}#compareTo in Dora, fixes 'unknown intrinsic FloatCmp/DoubleCmp'

soc:
Fixes https://github.com/dinfuehr/dora/issues/118.

It would be nicer to deal with this in assembly, but as long as I can't figure out what's wrong with `sortsAs` in assembly (https://github.com/dinfuehr/dora/compare/master...soc:topic/total-order?expand=1), this seems to be out of reach. So at least fix the crash for now.

### DIFF
--- a/dora/src/semck/prelude.rs
+++ b/dora/src/semck/prelude.rs
@@ -306,7 +306,6 @@ pub fn internal_functions<'ast>(vm: &mut VM<'ast>) {
     intrinsic_method(vm, clsid, "asInt", Intrinsic::FloatAsInt);
 
     intrinsic_method(vm, clsid, "equals", Intrinsic::FloatEq);
-    intrinsic_method(vm, clsid, "compareTo", Intrinsic::FloatCmp);
 
     intrinsic_method(vm, clsid, "plus", Intrinsic::FloatAdd);
     intrinsic_method(vm, clsid, "minus", Intrinsic::FloatSub);
@@ -328,7 +327,6 @@ pub fn internal_functions<'ast>(vm: &mut VM<'ast>) {
     intrinsic_method(vm, clsid, "asLong", Intrinsic::DoubleAsLong);
 
     intrinsic_method(vm, clsid, "equals", Intrinsic::DoubleEq);
-    intrinsic_method(vm, clsid, "compareTo", Intrinsic::DoubleCmp);
 
     intrinsic_method(vm, clsid, "plus", Intrinsic::DoubleAdd);
     intrinsic_method(vm, clsid, "minus", Intrinsic::DoubleSub);
--- a/dora/stdlib/Double.dora
+++ b/dora/stdlib/Double.dora
@@ -7,7 +7,13 @@
   @internal fun asLong() -> Long;
 
   @internal fun equals(rhs: Double) -> Bool;
-  @internal fun compareTo(rhs: Double) -> Int;
+  fun compareTo(rhs: Double) -> Int {
+    if self.isNan() || rhs.isNan() {
+      Int::minValue()
+    } else {
+      self.sortsAs(rhs)
+    }
+  }
   fun sortsAs(rhs: Double) -> Int {
     var ix = self.asLong();
     var iy = rhs.asLong();
--- a/dora/stdlib/Float.dora
+++ b/dora/stdlib/Float.dora
@@ -7,7 +7,13 @@
   @internal fun asInt() -> Int;
 
   @internal fun equals(rhs: Float) -> Bool;
-  @internal fun compareTo(rhs: Float) -> Int;
+  fun compareTo(rhs: Float) -> Int {
+    if self.isNan() || rhs.isNan() {
+      Int::minValue()
+    } else {
+      self.sortsAs(rhs)
+    }
+  }
   fun sortsAs(rhs: Float) -> Int {
     var ix = self.asInt();
     var iy = rhs.asInt();
--- a/tests/cmp.dora
+++ b/tests/cmp.dora
@@ -1,0 +1,85 @@
+fun main() {
+    assert(0.0F.compareTo(0.0F) == 0);
+    assert((-0.0F).compareTo(-0.0F) == 0);
+    assert(1.0F.compareTo(1.0F) == 0);
+    assert(0.0F.compareTo(1.0F) == -1);
+    assert(1.0F.compareTo(0.0F) == 1);
+    assert((-0.0F).compareTo(0.0F) == -1);
+    assert(0.0F.compareTo(-0.0F) == 1);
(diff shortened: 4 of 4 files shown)


soc:
```
failures:

---- bytecode::generator_tests::gen_expr_test_greaterthan_double stdout ----
thread 'bytecode::generator_tests::gen_expr_test_greaterthan_double' panicked at 'not implemented', dora/src/bytecode/generator.rs:915:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- bytecode::generator_tests::gen_expr_test_greaterthan_float stdout ----
thread 'bytecode::generator_tests::gen_expr_test_greaterthan_float' panicked at 'not implemented', dora/src/bytecode/generator.rs:915:13

---- bytecode::generator_tests::gen_expr_test_greaterthanequal_double stdout ----
thread 'bytecode::generator_tests::gen_expr_test_greaterthanequal_double' panicked at 'not implemented', dora/src/bytecode/generator.rs:915:13

---- bytecode::generator_tests::gen_expr_test_greaterthanequal_float stdout ----
thread 'bytecode::generator_tests::gen_expr_test_greaterthanequal_float' panicked at 'not implemented', dora/src/bytecode/generator.rs:915:13

---- bytecode::generator_tests::gen_expr_test_lessthan_float stdout ----
thread 'bytecode::generator_tests::gen_expr_test_lessthan_float' panicked at 'not implemented', dora/src/bytecode/generator.rs:915:13

---- bytecode::generator_tests::gen_expr_test_lessthan_double stdout ----
thread 'bytecode::generator_tests::gen_expr_test_lessthan_double' panicked at 'not implemented', dora/src/bytecode/generator.rs:915:13

---- bytecode::generator_tests::gen_expr_test_lessthanequal_double stdout ----
thread 'bytecode::generator_tests::gen_expr_test_lessthanequal_double' panicked at 'not implemented', dora/src/bytecode/generator.rs:915:13

---- bytecode::generator_tests::gen_expr_test_lessthanequal_float stdout ----
thread 'bytecode::generator_tests::gen_expr_test_lessthanequal_float' panicked at 'not implemented', dora/src/bytecode/generator.rs:915:13
```

What's the plan there?

soc:
@dinfuehr What's the best way to address this? Seems like the removal of `intrinsic_method(vm, clsid, "compareTo", Intrinsic::FloatCmp)` triggered this, thought this intrinsic was never implemented ...

dinfuehr:
It seems the intrinsic is used in baseline and cannon. Cannon assumes this to be an intrinsic, `visit_expr_bin` needs to change as well.

soc:
It's kinda weird, because `1 <= 2` gets treated as an intrinsic everywhere, while `1.compareTo(2)` does not – but removing `compareTo` from the list of intrinsics breaks `1 <= 2` in the bytecode.

dinfuehr:
I think Cannon doesn't support binary operators that are implemented by a regular method call.

soc:
> I think Cannon doesn't support binary operators that are implemented by a regular method call.

I don't think the binary operators are implemented by a regular method call – the `compareTo` impl doesn't have anything to do with `<`/`<=`/`>`/`>=` as far as I can see.

dinfuehr:
> > I think Cannon doesn't support binary operators that are implemented by a regular method call.
> 
> I don't think the binary operators are implemented by a regular method call – the `compareTo` impl doesn't have anything to do with `<`/`<=`/`>`/`>=` as far as I can see.

I think it does, there is some form of operator overloading happening here: https://github.com/dinfuehr/dora/blob/master/dora/src/typeck/expr.rs#L863-L902

Cannon already supports (most) function calls but it doesn't seem to be able to generate method calls from binary operators. Only if that method is backed by an intrinsic.

soc:
> I think it does, there is some form of operator overloading happening here: https://github.com/dinfuehr/dora/blob/master/dora/src/typeck/expr.rs#L863-L902

Thanks for the hint, I'll have a look!

dinfuehr:
I've talked with @hzahradnik about this, he mentioned he needs that specific feature for running benchmarks (String concatenation is a binary operator that expands to a method call), so he is going to work on this soon.

soc:
@dinfuehr Great, thanks! Then I'll wait and rebase this after the other changes are in.

Could it make sense to move toward an approach where we treat all operators as method calls as early as possible in the compilation phases?

dinfuehr:
> Could it make sense to move toward an approach where we treat all operators as method calls as early as possible in the compilation phases?

My first impression is not, but I might be missing something. What would this improve?

soc:
> My first impression is not, but I might be missing something. What would this improve?

It was just a hunch that we might be able to unify some codepaths, nothing important.

dinfuehr:
I think Cannon already improves the situation a lot: With Cannon the bytecode generator will handle most of that differences. While cannon/codegen.rs will be independent from the AST.

soc:
Thank you, good to know!

soc:
I rebased this, but it seems it worsens the output of the bytecode generator:

```
---- bytecode::generator_tests::gen_expr_test_greaterthan_float stdout ----
thread 'bytecode::generator_tests::gen_expr_test_greaterthan_float' panicked at 'assertion failed: `(left == right)`
  left: `[TestGtFloat(Register(2), Register(0), Register(1)), RetBool(Register(2))]`,
 right: `[PushRegister(Register(0)), PushRegister(Register(1)), InvokeStaticInt32(Register(3), FctDefId(0)), ConstInt32(Register(4), 0), TestGtInt32(Register(2), Register(3), Register(4)), RetBool(Register(2))]`', dora/src/bytecode/generator_tests.rs:1013:5
```
(8 fails in total)

I'd rather close this PR and focus on getting the intrinsics to work, such that need these workarounds in the first place, what do you think @dinfuehr?

dinfuehr:
This PR doesn't look too hard anymore, Cannon now successfully runs all tests. I think you could give it another chance.

soc:
True – though I realized the implementation using `sortsAs` is wrong. :-/

In addition to special-casing NaN we would need to also special-case 0.0 vs. -0.0.

In general the problem is that the `compareTo` needs to represent 4 states (smaller, bigger, equal and unordered), which also makes it a poor fit with the more specific <, <=, >, >= operations that return false in case of unordered.

One could make `compareTo` consistent with the more specific operations, but then I'm not sure it's a good idea. Though it would be nicer for integers,which otherwise would have unordered as a potential result (that is never actually returned).

soc:
@dinfuehr I'll give it another try after https://github.com/dinfuehr/dora/pull/205 by intrisifying `compareTo` similar to `sortsAs`, such that the operations don't have dependencies on each other.

PR #185, opened 2020-03-08
cannon: support for array

hzahradnik:
I started the implementation for array support in cannon. For now, only the allocation of a new Array is supported and only if the array is not a array of tuples, objects or unit, or is of length 0.

Before I continue, I thought it would be a good idea to get your feedback on this. Especially, I would like your opinion on my decision to change the order of how the arguments of a call are handled. The reason behind this decision is the length argument of the array, which is needed to calculate the size of the array. I could have just handled this case differently, but I see no reason, why this would be necessary.

Next, I would add functionality for set&get of array values, before I fill the missing types of arrays and the various edge cases (e.g. no length argument.)

### DIFF
--- a/dora/src/bytecode/generator.rs
+++ b/dora/src/bytecode/generator.rs
@@ -1,3 +1,4 @@
+use crate::semck::specialize::specialize_for_call_type;
 use dora_parser::lexer::position::Position;
 use std::collections::HashMap;
 
@@ -377,22 +378,30 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
             Register::zero()
         };
         self.gen.set_position(expr.pos);
+
         let arg_start_reg = match *call_type {
-            CallType::CtorNew(ty, _) => {
+            CallType::CtorNew(_, _) | CallType::Method(_, _, _) => 1,
+            _ => 0,
+        };
+
+        match *call_type {
+            CallType::CtorNew(_, _) => {
+                let ty = fct.params_with_self()[0];
+                let ty = specialize_for_call_type(&call_type, ty, self.vm);
+                let ty = self.specialize_type(ty);
                 let cls_id = specialize_class_ty(self.vm, ty);
+
                 self.gen.emit_new_object(start_reg, cls_id);
-                start_reg.offset(1)
             }
             CallType::Method(_, _, _) => {
                 let obj_expr = expr.object().expect("method target required");
                 self.visit_expr(obj_expr, DataDest::Reg(start_reg));
-                start_reg.offset(1)
             }
-            _ => start_reg,
+            _ => {}
         };
 
         for (idx, arg) in expr.args.iter().enumerate() {
-            let arg_reg = arg_start_reg.offset(idx);
+            let arg_reg = start_reg.offset(idx + arg_start_reg);
             self.visit_expr(arg, DataDest::Reg(arg_reg));
         }
         self.gen.set_position(expr.pos);
--- a/dora/src/bytecode/generator.rs
+++ b/dora/src/bytecode/generator.rs
@@ -10,6 +10,7 @@ use dora_parser::lexer::token::{FloatSuffix, IntSuffix};
 use crate::bytecode::{BytecodeFunction, BytecodeType, BytecodeWriter, Label, Register};
 use crate::semck::expr_block_always_returns;
 use crate::semck::specialize::{specialize_class_ty, specialize_type};
+use crate::size::InstanceSize;
 use crate::ty::{BuiltinType, TypeList};
 use crate::vm::{
     CallType, Fct, FctDef, FctDefId, FctId, FctKind, FctSrc, IdentType, Intrinsic, VarId, VM,
@@ -384,6 +385,11 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
             _ => 0,
         };
 
+        for (idx, arg) in expr.args.iter().enumerate() {
+            let arg_reg = start_reg.offset(idx + arg_start_reg);
+            self.visit_expr(arg, DataDest::Reg(arg_reg));
+        }
+
         match *call_type {
             CallType::CtorNew(_, _) => {
                 let ty = fct.params_with_self()[0];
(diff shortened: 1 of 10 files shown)


dora/src/bytecode/generator.rs
```diff
@@ -388,24 +390,47 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
+        };
+
+        for (idx, arg) in expr.args.iter().enumerate() {
+            let arg_reg = start_reg.offset(idx + arg_start_reg);
+            self.visit_expr(arg, DataDest::Reg(arg_reg));
+        }
+
+        match *call_type {
+            CallType::CtorNew(_, _) => {
+                let ty = fct.params_with_self()[0];
```
dinfuehr: I think you could take the object type out of `CtorNew` as well.

dinfuehr: Then you wouldn't need `specialize_for_call_type` here as well, since this type should already be specialized for that.

dora/src/bytecode/generator.rs
```diff
@@ -388,24 +390,47 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
             CallType::Method(_, _, _) => {
                 let obj_expr = expr.object().expect("method target required");
                 self.visit_expr(obj_expr, DataDest::Reg(start_reg));
-                start_reg.offset(1)
             }
-            _ => start_reg,
+            _ => {}
         };
 
-        for (idx, arg) in expr.args.iter().enumerate() {
```
dinfuehr: I think this would change the evaluation order of the arguments, consider the following program:

```
class Foo {
  println("Foo");
  fun bar(x: Int) { println("bar"); }
}
fun main() {
  Foo(1).bar(someInt())
}
fun someInt() {
  println("someInt");
}
```

Note this would be `CallType::Method` and not `CallType::Ctor`, so before this change we would have "Foo.someInt.bar", after it would print "someInt.Foo.bar". For `CallType::Method` we need to evaluate the first argument before the others. I guess we should add a test for this, such that we can ensure ast baseline works correctly and cannon matches this behavior.

On the other hand with `CallType::CtorNew` we can move `NewObject` after the arguments. Simply because that's guaranteed to be a simple object allocation (`NewObject` just allocates a raw object and doesn't run the constructor or any arbitrary code). Ast-baseline does this as well, I think the only reason for this is that the arguments could throw/fail and in that case we would have allocated the object for no reason. I think we could also allocate the object and array before the other arguments if this helps you.

On the other hand `visit_expr_call` already got quite large, it might make sense to split it:

1. arrays (we could make use of the fact there will only ever be one argument: length IIRC)
2. CtorNew (construct new object)
3. regular method calls (would probably also contain the parent ctor call)

I think it's better to have a bit of code duplication then to try to force everything through the same method and making the logic more and more complex. Especially with Cannon since generating bytecode is much easier.

I've realized that we already have baseline and cannon already differ in case the object is nil. Baseline will throw immediately and not evaluate the arguments, while cannon will first evaluate all arguments and then check if the object argument is nil. I think cannon is right here though and baseline should change.

dora/src/bytecode/generator.rs
hzahradnik: Before my change it was taken out of `CtorNew` and I'm pretty sure that it didn't work for classes with generics. But I will look into it again.

dora/src/bytecode/generator.rs
hzahradnik: > Note this would be `CallType::Method` and not `CallType::Ctor`, so before this change we would have "Foo.someInt.bar", after it would print "someInt.Foo.bar". For `CallType::Method` we need to evaluate the first argument before the others.

You're right, thanks for the example. I will change the evaluation order, so that the first argument is always evaluated before the other arguments, regardless of the call type.

>  I think we could also allocate the object and array before the other arguments if this helps you.

I think it is easier, if all arguments are evaluated in one go, but in the end, it doesn't really matter to me. Important for this PR is only, that the length argument is evaluated before `NewArray` is generated.

> On the other hand visit_expr_call already got quite large, it might make sense to split it:
> 1. arrays (we could make use of the fact there will only ever be one argument: length IIRC)
> 2. CtorNew (construct new object)
> 3. regular method calls (would probably also contain the parent ctor call)

I agree with you. After I'm done with this PR, I will create a new PR for this.

dora/src/bytecode/generator.rs
hzahradnik: I stand corrected, `CtorNew` has the correct type. Changed it back.

dora/src/bytecode/generator.rs
hzahradnik: Done. Added your example as a cannon test to assure the correct evaluation order.

hzahradnik:
I've implemented your suggestions. Is there something missing for this PR?

tests/cannon/evaluation-order1.dora
```diff
@@ -0,0 +1,15 @@
+//= stdout "Foo\nsomeInt\nbar\n"
```
dinfuehr: Can this test already be executed with Cannon? Maybe we could use `//= cannon` here

dora/src/bytecode/generator.rs
```diff
@@ -377,24 +378,46 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
+                        let length_arg = start_reg.offset(1);
+                        self.gen.emit_new_array(start_reg, cls_id, length_arg);
+                    }
+                    _ => {
+                        unimplemented!();
+                    }
+                }
+            }
+            _ => {}
+        };
```
dinfuehr: That semicolon seems to be superfluous

tests/cannon/evaluation-order1.dora
hzahradnik: I did't add `//= cannon`, but added `@cannon` to the methods. Otherwise it would break the "convention" of the tests in `tests/cannon`, which are all without the test annotation.
If it makes sense to run it twice with the baseline compiler and cannon, I could move it outside of the folder?

dora/src/bytecode/generator.rs
hzahradnik: It was. Removed it.

tests/cannon/evaluation-order1.dora
dinfuehr: Yes, I think it would make sense to run the test with both compilers. Just to ensure that their behavior matches.

dinfuehr:
Let's run that test on baseline as well and then merge this PR. Pretty cool work!

PR #186, opened 2020-03-08
Drop the `loop` keyword and replace usages with `while true`

soc:
This slightly reduces the amount of keywords/looping constructs/different ways to express the same thing.


soc:
This salvages the second commit of https://github.com/dinfuehr/dora/pull/100. (The `continue`/`break` change stays post-poned as you suggested.)

tests/loop1.dora
```diff
@@ -1,5 +1,23 @@
 fun main() {
+  println(x().toString());
+  // println(y().toString()); // terminated by signal SIGTRAP
```
dinfuehr: This bug is extremely bad. It is very likely that this isn't related to your changes: Either it was already there or I might have broken it recently when removing exceptions...

soc: Agreed.

dinfuehr: Ok, this is actually not a bug on master but only in this PR. I am a bit relieved now, was losing confidence in the test suite... The problem is that `while` loops are not always returning a value. This test case shouldn't even compile.

soc: As mentioned a few lines below in the comment, loop has the same bug:

```
  // the loop keyword had the same bug:
  /*
  loop {
    break;
    return 5;
  }
  */
```

(I agree that this bug should be fixed, but I believe that's unrelated to this change.)

dinfuehr: Oh right, thanks! Nevertheless handling of `while` is not quite correct. AFAICS this bug is in `semck` and should be gone after removing `loop`, so I guess I am waiting until this PR is ready.

soc: > AFAICS this bug is in semck and should be gone after removing loop, so I guess I am waiting until this PR is ready.

Not quite – I moved the functionality that was formerly on `loop` (returning things) to `while`, such that `while true` is a replacement for `loop` in all cases, not just in some, see: https://github.com/dinfuehr/dora/pull/186/files#diff-08481ad25efc97149a0a31c6823d17cf

> I guess I am waiting until this PR is ready.

Mhh, I'd say it is ready – what remains to be done in this PR from your POV?

dora/src/semck/returnck.rs
```diff
@@ -58,9 +58,8 @@ impl<'a, 'ast> Visitor<'ast> for ReturnCheck<'a, 'ast> {
 pub fn returns_value(s: &Stmt) -> Result<(), Position> {
     match *s {
         StmtReturn(_) => Ok(()),
-        StmtLoop(ref stmt) => returns_value(&stmt.block),
         StmtFor(ref stmt) => Err(stmt.pos),
-        StmtWhile(ref stmt) => Err(stmt.pos),
+        StmtWhile(ref stmt) => returns_value(&stmt.block),
```
dinfuehr: While this worked for `loop`, it doesn't work for `while` since the body might not be run. Just return `Err(stmt.pos)` here.

soc: > While this worked for loop, it doesn't work for while since the body might not be run.

It didn't work for `loop` either, see the code example. I'd argue that it should be made to work for `while`, just like it should have been fixed for `loop` if `loop` wasn't dropped.

> Just return Err(stmt.pos) here.

Will do!

soc: > Just return Err(stmt.pos) here.

Done!

dinfuehr: > It didn't work for `loop` either, see the code example. I'd argue that it should be made to work for `while` (require `return` after `while`), just like it should have been fixed for `loop` if `loop` wasn't dropped.
Well there was a bug in this analysis with loop, with `while` this isn't even allowed.

soc: I'd argue that _allowed & works_ > _allowed & buggy_ > _not allowed_.

Nevermind, let's deal with it step by step. :-)

dora/src/semck/returnck.rs
```diff
@@ -164,7 +162,7 @@ mod tests {
             pos(1, 16),
             SemError::ReturnType("Int".into(), "()".into()),
         );
-        ok("fun f() -> Int { loop { return 1; } }");
+        // ok("fun f() -> Int { while true { return 1; } }");
```
dinfuehr: This should be an error

tests/loop1.dora
dinfuehr: I think you can remove that comment now.

tests/loop1.dora
```diff
@@ -1,5 +1,28 @@
...
+
+fun x() -> Int {
+  while true {
+    break;
+  }
+  return 1;
+}
+
+fun y() -> Int {
+  // should not compile; should require return after the while loop:
```
dinfuehr: I am not sure we need this test at all, I guess it is best to just delete it.

soc:
@dinfuehr Updated!

dinfuehr:
Thanks!

PR #187, opened 2020-03-11
Fix the error messages for annotation parsing problems

soc:
The keyword was printed instead of the offending annotation, e. g.:

    @open @open
    class Foo()

Before: redundant annotation class
After:  redundant annotation open


dinfuehr:
Thanks!

PR #192, opened 2020-03-23
Split terms and types into separate namespaces

soc:
This reduces the amount of "surprising" name clashes¹ and
is required for both modules and annotations to work well:

- It allows modules to have the same name as a class or struct.
  This is a stepping stone to replacing @static members with members inside modules.
- It avoids annotation names clashing with function names (like in `fun test()`).¹

¹ Annotations (which are types) start with a lowercase letter because it looks more familiar/
  more like a modifier and draws away less attention from the keywords that follows.
  Nevertheless, we get some additional benefit out of it:
  It makes annotations unlikely to ever clash with another type definition,
  because classes, traits, structs all start with an uppercase letter by convention.

### DIFF
--- a/dora/src/semck/globaldef.rs
+++ b/dora/src/semck/globaldef.rs
@@ -4,10 +4,11 @@ use std::sync::Arc;
 
 use crate::error::msg::SemError;
 use crate::gc::Address;
-use crate::sym::Sym::{
-    self, SymClass, SymConst, SymEnum, SymFct, SymGlobal, SymModule, SymStruct, SymTrait,
+use crate::sym::TermSym::{
+    SymClassConstructor, SymConst, SymFct, SymGlobal, SymModule, SymStructConstructor, SymVar,
 };
-use crate::sym::SymLevel;
+use crate::sym::TypeSym::{SymClass, SymEnum, SymStruct, SymTrait};
+use crate::sym::{SymLevel, TermSym, TypeSym};
 use crate::ty::BuiltinType;
 use crate::vm::module::ModuleId;
 use crate::vm::{
@@ -81,12 +82,12 @@ impl<'x, 'ast> Visitor<'ast> for GlobalDef<'x, 'ast> {
         };
 
         self.vm.traits.push(RwLock::new(xtrait));
-        let sym = SymTrait(id);
 
         self.map_trait_defs.insert(t.id, id);
 
-        if let Some(sym) = self.vm.sym.lock().insert(t.name, sym) {
-            report(self.vm, t.name, self.file_id.into(), t.pos, sym);
+        let sym = SymTrait(id);
+        if let Some(sym) = self.vm.sym.lock().insert_type(t.name, sym) {
+            report_type_shadow(self.vm, t.name, self.file_id.into(), t.pos, sym);
         }
     }
 
@@ -111,11 +112,11 @@ impl<'x, 'ast> Visitor<'ast> for GlobalDef<'x, 'ast> {
             id
         };
 
-        let sym = SymGlobal(id);
         self.map_global_defs.insert(g.id, id);
 
-        if let Some(sym) = self.vm.sym.lock().insert(g.name, sym) {
-            report(self.vm, g.name, self.file_id.into(), g.pos, sym);
+        let sym = SymGlobal(id);
+        if let Some(sym) = self.vm.sym.lock().insert_term(g.name, sym) {
+            report_term_shadow(self.vm, g.name, self.file_id.into(), g.pos, sym);
         }
     }
 
@@ -178,12 +179,11 @@ impl<'x, 'ast> Visitor<'ast> for GlobalDef<'x, 'ast> {
             id
         };
 
-        let sym = SymModule(id);
-
         self.map_module_defs.insert(m.id, id);
 
-        if let Some(sym) = self.vm.sym.lock().insert(m.name, sym) {
-            report(self.vm, m.name, self.file_id.into(), m.pos, sym);
+        let sym = SymModule(id);
+        if let Some(sym) = self.vm.sym.lock().insert_term(m.name, sym) {
+            report_term_shadow(self.vm, m.name, self.file_id.into(), m.pos, sym);
         }
     }
 
@@ -209,9 +209,8 @@ impl<'x, 'ast> Visitor<'ast> for GlobalDef<'x, 'ast> {
         self.map_const_defs.insert(c.id, id);
 
         let sym = SymConst(id);
-
-        if let Some(sym) = self.vm.sym.lock().insert(c.name, sym) {
-            report(self.vm, c.name, self.file_id.into(), c.pos, sym);
+        if let Some(sym) = self.vm.sym.lock().insert_term(c.name, sym) {
+            report_term_shadow(self.vm, c.name, self.file_id.into(), c.pos, sym);
         }
     }
 
@@ -262,12 +261,20 @@ impl<'x, 'ast> Visitor<'ast> for GlobalDef<'x, 'ast> {
             id
         };
 
-        let sym = SymClass(id);
-
         self.map_cls_defs.insert(c.id, id);
 
-        if let Some(sym) = self.vm.sym.lock().insert(c.name, sym) {
-            report(self.vm, c.name, self.file_id.into(), c.pos, sym);
+        let sym = SymClass(id);
+        if let Some(sym) = self.vm.sym.lock().insert_type(c.name, sym) {
+            report_type_shadow(self.vm, c.name, self.file_id.into(), c.pos, sym);
+            return;
+        }
+
+        let sym = SymClassConstructor(id);
+        match self.vm.sym.lock().insert_term(c.name, sym) {
+            Some(SymModule(_)) | None => {}
+            Some(sym) => {
+                report_term_shadow(self.vm, c.name, self.file_id.into(), c.pos, sym);
+            }
         }
     }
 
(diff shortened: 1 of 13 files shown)


soc:

dinfuehr:
Will take a look on the weekend...

soc:
@dinfuehr Thanks!

soc:
When you are happy with this, I have another PR queued up that implements calling funs in modules, and an additional one which migrates the standard library and updates the prelude accordingly.

dora/src/semck/globaldef.rs
```diff
@@ -401,11 +424,7 @@ mod tests {
             pos(1, 15),
             SemError::ShadowStruct("Foo".into()),
         );
-        err(
-            "struct Foo {} fun Foo() {}",
-            pos(1, 15),
-            SemError::ShadowStruct("Foo".into()),
-        );
+        ok("struct Foo {} fun Foo() {}");
```
dinfuehr: Not sure this should be allowed: we would allow `class Foo {} fun Foo() {}`as well and then we don't know from `Foo()` whether we we should invoke the constructor/object or the function.

dinfuehr: Oh it would be the same for globals and vars as well, e.g. from `Foo(1)` we don't know whether we want to create an object or access an array element.

soc: You are right, I didn't intend to exclude struct; also right on the rest – I'll fix it. Thanks!

soc: @dinfuehr I pushed an updated PR.

dora/src/sym.rs
```diff
@@ -44,118 +55,111 @@ impl SymTable {
...
+
+    pub fn insert_term(&mut self, name: Name, sym: TermSym) -> Option<TermSym> {
+        self.levels.last_mut().unwrap().insert_term(name, sym)
     }
 }
 
 #[derive(Debug)]
 pub struct SymLevel {
-    map: HashMap<Name, Sym>,
+    types: HashMap<Name, TypeSym>,
```
dinfuehr: While I like this split between `types` and `terms`, I am not convinced this is the right thing to do. AFAIU for example functions and classes live in separate maps, however we are not going to allow having both a function and class with the same name. The split doesn't really match language semantics.

Why not simply introduce separate maps for both modules and annotations instead?

soc: > AFAIU for example functions and classes live in separate maps, however we are not going to allow having both a function and class with the same name. The split doesn't really match language semantics.

I actually think it does – it's just that the way I'm checking it currently is clumsy:

When defining a `class Foo` (or struct) we should consider the name to be in both namespaces  (in _types_ because a class is a type; and in _terms_ because the class implicitly defines a  constructor that could be seen as a "static method").

Thinking about it, maybe I should just do exactly this – enter the name in both namespaces when encountering a class definition – then we can get rid of the special casing altogether.

> Why not simply introduce separate maps for both modules and annotations instead?

I think it's not a good idea to introduce ad-hoc namespaces – because these things tend to keep accumulating (Java has like 7 now) and make many things more complicated later on.

I believe the division between terms and types is a good choice based on languages that have done this (I mean both splitting things like this and the number of namespaces (2)).

dinfuehr: > > AFAIU for example functions and classes live in separate maps, however we are not going to allow having both a function and class with the same name. The split doesn't really match language semantics.
> 
> I actually think it does – it's just that the way I'm checking it currently is clumsy:
> 
> When defining a `class Foo` (or struct) we should consider the name to be in both namespaces (in _types_ because a class is a type; and in _terms_ because the class implicitly defines a constructor that could be seen as a "static method").
> 
> Thinking about it, maybe I should just do exactly this – enter the name in both namespaces when encountering a class definition – then we can get rid of the special casing altogether.

This sounds good, maybe then we can also mostly keep the current API and can handle the difference internally in `SymLevel`. For example I would probably still not split `Sym`, but depending on the `Sym` kind insert into a different map (or multiple).

> 
> > Why not simply introduce separate maps for both modules and annotations instead?
> 
> I think it's not a good idea to introduce ad-hoc namespaces – because these things tend to keep accumulating (Java has like 7 now) and make many things more complicated later on.
> 
> I believe the division between terms and types is a good choice based on languages that have done this (I mean both splitting things like this and the number of namespaces (2)).

I guess it depends on what kinds of overlaps you want to allow. But I agree that being more restrictive here is better.

soc: > This sounds good ...

Thanks for making me think about it, I think this solution will be way easier to read in the code.

> ... maybe then we can also mostly keep the current API and can handle the difference internally in SymLevel

I tried this before settling on splitting the symbols – I tried both having things as term-type-tuples and having Syms that are multiple things at once.

The outcomes of these approaches were not great complexity-wise.

I'd try to keep the split, because I believe that in the future we will have more places where we benefit from more matches that don't need to have catch-all case like in https://github.com/dinfuehr/dora/pull/192/files#diff-86db169ca86f2fbc58c5deec22c19fe0L325 or https://github.com/dinfuehr/dora/pull/192/files#diff-86db169ca86f2fbc58c5deec22c19fe0L455.

> I guess it depends on what kinds of overlaps you want to allow. But I agree that being more restrictive here is better.

Agreed, I'd try to be as restrictive as possible without it being an annoyance to write code.
There is a lot of benefit to be had of being able to say "I can look at this and know immediately to what this name refers to".

soc: I pushed changes to get rid of the special-handling in all places except when entering the class name as a constructor itself.

I think doing it this way is an improvemen, what do you think @dinfuehr?

dora/src/semck/globaldef.rs
```diff
@@ -398,15 +459,40 @@ mod tests {
...
+        err(
+            "fun Foo() {} struct Foo {}",
+            pos(1, 14),
+            SemError::ShadowFunction("Foo".into()),
+        );
         err(
             "struct Foo {} fun Foo() {}",
             pos(1, 15),
-            SemError::ShadowStruct("Foo".into()),
+            SemError::ShadowStructConstructor("Foo".into()),
```
dinfuehr, line 480: Not sure we need ShadowStructConstructor right now, structs are not implemented yet but in the initial implementation would be intialized like this `Foo { .. }`. So from this we could see that this is a type and not a term. Wdyt?

dora/src/sym.rs
dinfuehr: Indeed this is now much cleaner 👍

dora/src/semck/globaldef.rs
```diff
@@ -390,6 +417,40 @@ mod tests {
...
+            pos(1, 14),
+            SemError::ShadowFunction("Foo".into()),
+        );
+        err(
+            "class Foo fun Foo() {}",
+            pos(1, 11),
+            SemError::ShadowClassConstructor("Foo".into()),
+        );
+        err(
+            "class Foo let Foo: Int = 1;",
```
dinfuehr, line 438: Could you please add a test case for a local variable as well?

dora/src/semck/globaldef.rs
soc: I would avoid having multiple different syntax variations that do the same thing (I would recommend that structs are initialized with `Foo(..)` and letting an IDE make the distinction between class and struct by coloring things differently.

Rust is a good example where giving initialization and calls different syntax has worked out quite poorly, I'd really try to avoid repeating those mistakes.

dora/src/semck/globaldef.rs
soc: You mean like

```
class Foo {
    var Foo: Foo = nil;
}
```

or something different?

dora/src/sym.rs
soc: Great!

(I'm hoping that I can get rid of the remaining special-hanbdling as well in the mid-to-long-term by treating the constructor more as a first-class element.)

dora/src/semck/globaldef.rs
soc: @dinfuehr I added some checks here: https://github.com/dinfuehr/dora/pull/192/files#diff-313b446e400b66c47affefe1a6b19787R619

dora/src/semck/globaldef.rs
dinfuehr: Using `Foo(..)` would definitely make parsing easier. However we then really have to use named arguments at least in those cases: `Point(x: 1, y: 2)`.

soc: Agreed. I think named argument should not be special-case that works only within curly braces for structs like it does in Rust.

Instead, parameter lists should have named arguments -> constructors are parameter lists -> struct initializaters are constructors.

soc:
@dinfuehr Is this good to go?

dora/src/semck/globaldef.rs
soc: Regarding named arguments: I would go with an intentionally constraint design:

- Params can be named or unnamed.
- Do not allow reordering of params.

This gets us 95% of the benefits and avoids many gnarly problems I experienced in the past.

dora/src/semck/globaldef.rs
dinfuehr: Yes this PR is almost good to go. I think we should have a test for clashes between class name and local variable as well: "class Foo fun x() { let Foo: Int = 1; }". AFAIU this is allowed, but I think it would to add an explicit test for that.

soc: I'm on it! Give me 5 minutes...

soc: It wasn't allowed – I added a commit that allows it: https://github.com/dinfuehr/dora/pull/192/commits/25671411cd63c563d871c3bf0e9bc8c31a439dd9

soc: If you are able to merge this, how long are you available after that? If you are still here for a while, I'd rebase the other PRs on the merged one and try to get them through, too.

soc:
@dinfuehr Thanks! I'll rebase the other PRs on master now!

PR #193, opened 2020-03-29
Module methods

soc:
This means that

```
module Foo {
  fun foo() -> String = "module";
  fun bar() -> String = Foo().foo();
  fun baz() -> String = Foo::foo();
}

fun main() {
  assert(Foo::foo() == "module");
  assert(Foo::bar() == "class");
  assert(Foo::baz() == "module");
}
```

works now.

The only remaining usages of `@static` in the standard library are in the
`Default` and `Zero` traits – I have a concept for them too, but this requires
additional preparations in other places, which exceed the scope of this PR.

### DIFF
--- a/dora/src/baseline/codegen.rs
+++ b/dora/src/baseline/codegen.rs
@@ -1342,6 +1342,7 @@ where
             &IdentType::Fct(_) | &IdentType::FctType(_, _) => unreachable!(),
             &IdentType::Class(_) | &IdentType::ClassType(_, _) => unreachable!(),
             &IdentType::Module(_) => unreachable!(),
+            &IdentType::ClassAndModule(_, _) => unreachable!(),
             &IdentType::Method(_, _) | &IdentType::MethodType(_, _, _) => unreachable!(),
             &IdentType::TypeParam(_) | &IdentType::TypeParamStaticMethod(_, _) => unreachable!(),
             &IdentType::StaticMethod(_, _) | &IdentType::StaticMethodType(_, _, _) => {
@@ -1716,6 +1717,7 @@ where
             &IdentType::Fct(_) | &IdentType::FctType(_, _) => unreachable!(),
             &IdentType::Class(_) | &IdentType::ClassType(_, _) => unreachable!(),
             &IdentType::Module(_) => unreachable!(),
+            &IdentType::ClassAndModule(_, _) => unreachable!(),
             &IdentType::Method(_, _) | &IdentType::MethodType(_, _, _) => unreachable!(),
             &IdentType::TypeParam(_) | &IdentType::TypeParamStaticMethod(_, _) => unreachable!(),
             &IdentType::StaticMethod(_, _) | &IdentType::StaticMethodType(_, _, _) => {
@@ -1963,6 +1965,8 @@ where
                         fct_id
                     }
                 }
+
+                CallType::ModuleMethod(_, fct_id, _) => fct_id,
 
                 CallType::Fct(fid, _, _) => fid,
 
@@ -3639,6 +3643,13 @@ where
             }
 
             CallType::Method(ty, _, ref type_params) => {
+                let ty = self.specialize_type(ty);
+
+                cls_type_params = ty.type_params(self.vm);
+                fct_type_params = type_params.clone();
+            }
+
+            CallType::ModuleMethod(ty, _, ref type_params) => {
                 let ty = self.specialize_type(ty);
 
                 cls_type_params = ty.type_params(self.vm);
--- a/dora/src/bytecode/generator.rs
+++ b/dora/src/bytecode/generator.rs
@@ -779,7 +779,7 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
 
         // self was already emitted, needs to be ignored here.
         let arg_start_offset = match *call_type {
-            CallType::CtorNew(_, _) | CallType::Method(_, _, _) | CallType::Expr(_, _) => 1,
+            CallType::CtorNew(_, _) | CallType::Expr(_, _) | CallType::Method(_, _, _) => 1,
             _ => 0,
         };
 
@@ -916,6 +916,13 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
                 } else if fct.is_virtual() {
                     self.emit_invoke_virtual(return_type, return_reg, fct_def_id);
                 } else if arg_bytecode_types[0] != BytecodeType::Ptr {
+                    self.emit_invoke_static(return_type, return_reg, fct_def_id);
+                } else {
+                    self.emit_invoke_direct(return_type, return_reg, fct_def_id);
+                }
+            }
+            CallType::ModuleMethod(_, _, _) => {
+                if arg_bytecode_types.is_empty() || arg_bytecode_types[0] != BytecodeType::Ptr {
                     self.emit_invoke_static(return_type, return_reg, fct_def_id);
                 } else {
                     self.emit_invoke_direct(return_type, return_reg, fct_def_id);
@@ -2221,6 +2228,7 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
             &IdentType::Fct(_) | &IdentType::FctType(_, _) => unreachable!(),
             &IdentType::Class(_) | &IdentType::ClassType(_, _) => unreachable!(),
             &IdentType::Module(_) => unreachable!(),
+            &IdentType::ClassAndModule(_, _) => unreachable!(),
             &IdentType::TypeParam(_) | &IdentType::TypeParamStaticMethod(_, _) => unreachable!(),
             &IdentType::Method(_, _) | &IdentType::MethodType(_, _, _) => unreachable!(),
             &IdentType::StaticMethod(_, _) | &IdentType::StaticMethodType(_, _, _) => {
@@ -2406,6 +2414,13 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
                 fct_type_params = type_params.clone();
             }
 
+            CallType::ModuleMethod(ty, _, ref type_params) => {
+                let ty = self.specialize_type(ty);
+
+                cls_type_params = ty.type_params(self.vm);
+                fct_type_params = type_params.clone();
+            }
+
             CallType::Fct(_, ref cls_tps, ref fct_tps) => {
                 cls_type_params = cls_tps.clone();
                 fct_type_params = fct_tps.clone();
@@ -2458,6 +2473,11 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
             }
 
             CallType::Method(cls_ty, _, ref type_params) => {
+                let cls_type_params = cls_ty.type_params(self.vm);
+                specialize_type(self.vm, ty, &cls_type_params, type_params)
+            }
+
+            CallType::ModuleMethod(cls_ty, _, ref type_params) => {
                 let cls_type_params = cls_ty.type_params(self.vm);
                 specialize_type(self.vm, ty, &cls_type_params, type_params)
             }
--- a/dora/src/semck/nameck.rs
+++ b/dora/src/semck/nameck.rs
@@ -187,6 +187,12 @@ impl<'a, 'ast> NameCheck<'a, 'ast> {
         let type_sym = self.vm.sym.lock().get_type(ident.name);
 
         match (term_sym, type_sym) {
+            (Some(SymModule(module_id)), Some(SymClass(class_id))) => {
+                self.src
+                    .map_idents
+                    .insert(ident.id, IdentType::ClassAndModule(class_id, module_id));
+            }
+
             (Some(SymVar(id)), None) => {
                 self.src.map_idents.insert(ident.id, IdentType::Var(id));
             }
--- a/dora/src/typeck/expr.rs
+++ b/dora/src/typeck/expr.rs
@@ -421,7 +421,7 @@ impl<'a, 'ast> TypeCheck<'a, 'ast> {
                 BuiltinType::Error
             }
 
-            &IdentType::Module(module_id) => {
+            &IdentType::Module(module_id) | &IdentType::ClassAndModule(_, module_id) => {
                 let module = self.vm.modules.idx(module_id);
                 let ty = module.read().ty;
                 self.src.set_ty(e.id, ty);
@@ -554,6 +554,8 @@ impl<'a, 'ast> TypeCheck<'a, 'ast> {
             }
 
             &IdentType::Module(_) => unreachable!(),
+
+            &IdentType::ClassAndModule(_, _) => unreachable!(),
 
             &IdentType::TypeParam(_) | &IdentType::TypeParamStaticMethod(_, _) => {
                 self.vm
@@ -991,7 +993,7 @@ impl<'a, 'ast> TypeCheck<'a, 'ast> {
                 self.check_expr_call_ident(e, fct_id, type_params, &arg_types)
             }
 
-            Some(IdentType::Class(cls_id)) => {
+            Some(IdentType::Class(cls_id)) | Some(IdentType::ClassAndModule(cls_id, _)) => {
                 self.check_expr_call_ctor(e, cls_id, TypeList::empty(), &arg_types)
             }
 
@@ -1336,7 +1338,11 @@ impl<'a, 'ast> TypeCheck<'a, 'ast> {
                 CallType::Trait(trait_id, fct_id)
             } else {
                 let method_type = lookup.found_class_type().unwrap();
-                CallType::Method(method_type, fct_id, type_params.clone())
+                if method_type.is_module() {
+                    CallType::ModuleMethod(method_type, fct_id, type_params.clone())
+                } else {
+                    CallType::Method(method_type, fct_id, type_params.clone())
+                }
             };
 
             self.src
@@ -1628,7 +1634,8 @@ impl<'a, 'ast> TypeCheck<'a, 'ast> {
                 IdentType::StaticMethod(cls_ty, name)
             }
 
-            Some(&IdentType::Module(module_id)) => {
+            Some(&IdentType::Module(module_id))
+            | Some(&IdentType::ClassAndModule(_, module_id)) => {
                 let module_ty = BuiltinType::Module(module_id);
 
                 IdentType::Method(module_ty, name)
@@ -1722,7 +1729,7 @@ impl<'a, 'ast> TypeCheck<'a, 'ast> {
         let type_params: TypeList = TypeList::with(type_params);
 
         match ident_type {
-            Some(IdentType::Class(cls_id)) => {
+            Some(IdentType::Class(cls_id)) | Some(IdentType::ClassAndModule(cls_id, _)) => {
                 self.src
                     .map_idents
                     .insert(e.id, IdentType::ClassType(cls_id, type_params));
--- a/dora/src/typeck/lookup.rs
+++ b/dora/src/typeck/lookup.rs
@@ -292,7 +292,6 @@ impl<'a, 'ast> MethodLookup<'a, 'ast> {
                 let list_id = self.vm.lists.lock().insert(cls_tps);
                 BuiltinType::Class(cls_id, list_id)
             }
-
             _ => replace_type_param(self.vm, fct.return_type, &cls_tps, &fct_tps, None),
         };
 
--- a/dora/src/typeck/tests.rs
+++ b/dora/src/typeck/tests.rs
@@ -70,22 +70,56 @@ fn type_class_method_call() {
 }
 
 #[test]
-fn type_module_method_call_1() {
+fn type_module_method_call() {
     ok("module Foo {
                 fun bar() {}
                 fun baz() -> Int32 { return 1; }
             }
 
-            fun f1() { Foo.bar(); }
+            class Foo() {}
+
             fun f2() { Foo::bar(); }
-            fun g() -> Int32 { return Foo.baz(); }");
+            fun g() -> Int32 { return Foo::baz(); }");
+
+    ok("module Foo {
+                fun bar[T : Equals + Hash](t: T) -> T = t;
+            }
+
+            fun foo() { Foo::bar[Int32](1); }");
+
+    err(
+        "module Foo {
+                 fun foo() -> Bar = Bar();
+               }
+               class Bar[T]",
+        pos(2, 31),
+        SemError::WrongNumberTypeParams(1, 0),
+    );
+
+    err(
+        "module Foo {
+                 fun foo[T]() -> Bar = Bar[T]();
+               }
+               class Bar[T]",
+        pos(2, 34),
+        SemError::WrongNumberTypeParams(1, 0),
+    );
+
+    ok("module Foo {
+                 fun foo[T]() -> Bar[T] = Bar[T]();
+               }
+               class Bar[T]
+
+               fun bar() {
+                 Foo::foo[Int32]();
+               }");
 
     err(
         "module Foo {
                  fun bar() -> Int32 { return 0; }
              }
 
-             fun f() -> String { return Foo.bar(); }",
+             fun f() -> String { return Foo::bar(); }",
         pos(5, 34),
         SemError::ReturnType("String".into(), "Int32".into()),
     );
@@ -606,11 +640,7 @@ fn access_super_class_field() {
 fn same_names() {
     ok("class Foo { var Foo: Foo = Foo(); }");
     ok("class Foo fun foo() { let Foo: Int32 = 1; }");
-    err(
-        "class Foo { var Foo: Foo = Foo(); } module Foo { fun Foo() -> Foo = nil; }",
-        pos(1, 37),
-        SemError::ShadowClassConstructor("Foo".into()),
-    );
+    ok("class Foo { fun Foo() -> Foo { Foo::Foo(); return Foo(); } } module Foo { fun Foo() -> Foo = nil; }");
 }
 
 #[test]
--- a/dora/src/vm/src.rs
+++ b/dora/src/vm/src.rs
@@ -164,8 +164,11 @@ pub enum IdentType {
     // name of class with type params: SomeClass[T1, T2, ...]
     ClassType(ClassId, TypeList),
 
-    // name of method
+    // name of module
     Module(ModuleId),
+
+    // name of class and module
+    ClassAndModule(ClassId, ModuleId),
 
     // method expression: <expr>.<method_name>
     Method(BuiltinType, Name),
@@ -254,6 +257,9 @@ pub enum CallType {
     // Direct or virtual method calls, e.g. obj.method(<args>)
     Method(BuiltinType, FctId, TypeList),
 
+    // Module method call, e.g. Module::method(<args>)
+    ModuleMethod(BuiltinType, FctId, TypeList),
+
     // Constructor call Class(<args>)
     CtorNew(BuiltinType, FctId),
     Ctor(BuiltinType, FctId),
@@ -311,6 +317,7 @@ impl CallType {
         match *self {
             CallType::Fct(fctid, _, _) => Some(fctid),
             CallType::Method(_, fctid, _) => Some(fctid),
+            CallType::ModuleMethod(_, fctid, _) => Some(fctid),
             CallType::CtorNew(_, fctid) => Some(fctid),
             CallType::Ctor(_, fctid) => Some(fctid),
             CallType::Expr(_, fctid) => Some(fctid),
--- a/tests/module/module2.dora
+++ b/tests/module/module2.dora
@@ -1,0 +1,9 @@
+module Foo {
+  fun foo[T]() -> Bar[T] = Bar[T](23);
+}
+
+class Bar[T](let x: Int32)
+
+fun main() {
+  Foo::foo[Int32]();
+}
--- a/dora/src/semck/clsdefck.rs
+++ b/dora/src/semck/clsdefck.rs
@@ -317,8 +317,7 @@ impl<'x, 'ast> Visitor<'ast> for ClsDefCheck<'x, 'ast> {
             has_optimize_immediately: f.has_optimize_immediately,
             variadic_arguments: false,
 
-            // abstract for methods also means that method is open to
-            // override
+            // abstract for methods also means that method is open to override
             has_open: f.has_open || f.is_abstract,
             has_final: f.has_final,
             is_pub: f.is_pub,
--- a/dora/src/semck/fctdefck.rs
+++ b/dora/src/semck/fctdefck.rs
@@ -186,6 +186,12 @@ pub fn check<'a, 'ast>(vm: &VM<'ast>) {
                 check_against_methods(vm, &*fct, &ximpl.methods);
             }
 
+            FctParent::Module(modid) => {
+                let module = vm.modules.idx(modid);
+                let module = module.read();
+                check_against_methods(vm, &*fct, &module.methods);
+            }
+
             _ => {}
         }
 
--- a/dora/src/semck/globaldef.rs
+++ b/dora/src/semck/globaldef.rs
@@ -182,8 +182,14 @@ impl<'x, 'ast> Visitor<'ast> for GlobalDef<'x, 'ast> {
         self.map_module_defs.insert(m.id, id);
 
         let sym = SymModule(id);
-        if let Some(sym) = self.vm.sym.lock().insert_term(m.name, sym) {
-            report_term_shadow(self.vm, m.name, self.file_id.into(), m.pos, sym);
+        let term_sym = self.vm.sym.lock().get_term(m.name);
+        match term_sym {
+            None | Some(SymClassConstructor(_)) => {
+                self.vm.sym.lock().insert_term(m.name, sym);
+            }
+            Some(sym) => {
+                report_term_shadow(self.vm, m.name, self.file_id.into(), m.pos, sym);
+            }
         }
     }
 
@@ -270,8 +276,12 @@ impl<'x, 'ast> Visitor<'ast> for GlobalDef<'x, 'ast> {
         }
 
         let sym = SymClassConstructor(id);
-        match self.vm.sym.lock().insert_term(c.name, sym) {
-            Some(SymModule(_)) | None => {}
+        let term_sym = self.vm.sym.lock().get_term(c.name);
+        match term_sym {
+            Some(SymModule(_)) => {}
+            None => {
+                self.vm.sym.lock().insert_term(c.name, sym);
+            }
             Some(sym) => {
                 report_term_shadow(self.vm, c.name, self.file_id.into(), c.pos, sym);
             }
@@ -305,8 +315,12 @@ impl<'x, 'ast> Visitor<'ast> for GlobalDef<'x, 'ast> {
         }
 
         let sym = SymStructConstructor(id);
-        match self.vm.sym.lock().insert_term(s.name, sym) {
-            Some(SymModule(_)) | None => {}
+        let term_sym = self.vm.sym.lock().get_term(s.name);
+        match term_sym {
+            Some(SymModule(_)) => {}
+            None => {
+                self.vm.sym.lock().insert_term(s.name, sym);
+            }
             Some(sym) => {
                 report_term_shadow(self.vm, s.name, self.file_id.into(), s.pos, sym);
             }
--- a/dora/src/semck/moduledefck.rs
+++ b/dora/src/semck/moduledefck.rs
@@ -341,6 +341,11 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_module_fun_call() {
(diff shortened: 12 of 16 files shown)


soc:
I rebased on the latest changes in #192.

soc:

soc:
Rebased.

soc:
@dinfuehr test failure is new after rebase:

```
topic/module-funs ⨯ RUST_BACKTRACE=1 target/debug/dora tests/boots/hello.dora                                                                                                                                                                           10:49:44
thread 'main' panicked at 'missing method compile', dora/src/boots.rs:29:26
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1052
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1428
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:204
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:224
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:472
  11: rust_begin_unwind
             at src/libstd/panicking.rs:380
  12: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  13: core::option::expect_failed
             at src/libcore/option.rs:1191
  14: core::option::Option<T>::expect
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/option.rs:347
  15: dora::boots::compile
             at dora/src/boots.rs:29
  16: dora::compiler::codegen::generate_fct
             at dora/src/compiler/codegen.rs:70
  17: dora::compiler::codegen::generate
             at dora/src/compiler/codegen.rs:30
  18: dora::compiler::compile_stub::patch_fct_call
             at dora/src/compiler/compile_stub.rs:258
  19: dora::compiler::compile_stub::compile_request
             at dora/src/compiler/compile_stub.rs:219
```

Any ideas?

dinfuehr:
Not sure, somehow the `compile` method (defined in dora-boots/compile.dora) is missing/not found. It seems to be working on ToT, so I guess your PR is changing something here.

soc:
ToT?

dinfuehr:
Sorry, ToT=tip of tree

soc:
@dinfuehr I just checked out current master – the test fails on my machine at ToT. Any ideas?

I cleaned and rebuilt completely, don't have any stale files, etc.

dinfuehr:
Hmm, that's strange. It seems to work for me on multiple machines and on all the CI bots. Are you running this test manually? If yes, than you need to pass an additional argument: `--boots=dora-boots`. Oh right, it seems that the argument is missing in the command line invocation above.

dinfuehr:
You might want to run `tools/test-boots` as well. This runs only the boots-related tests.

soc:
@dinfuehr Fixed it, silly mistake on my side, adding `--boots=dora-boots` immediately showed the issue. Thanks!

soc:
Rebased, everything works except `tests/boots/hello.dora`:

```
thread 'main' panicked at 'no module', dora/src/vm/module.rs:74:25
```

Debugging tells me it's looking for `uint8` when failing to find the module ...

dinfuehr:
The tests are passing on all bots, so I would assume the failure is related to this PR. That said, could we in a first PR introduce the concept and in a second PR then update all usages? I think this is one of the things we could do to land PRs faster.

soc:
@dinfuehr Done. This PR now only includes the concept and the changes to the standard library. I'll do a second PR for the remaining changes. Thanks!

dora-parser/src/ast/dump.rs
```diff
@@ -624,15 +624,15 @@ impl<'a> AstDumper<'a> {
     }
 
     fn dump_expr_dot(&mut self, expr: &ExprDotType) {
-        self.indent(|d| d.dump_expr(&expr.rhs));
         dump!(self, "dot @ {} {}", expr.pos, expr.id);
         self.indent(|d| d.dump_expr(&expr.lhs));
+        self.indent(|d| d.dump_expr(&expr.rhs));
```
dinfuehr, line 629: Why this change? This should be the same as e.g. in dump_expr_bin.

dora-parser/src/ast/dump.rs
```diff
@@ -624,15 +624,15 @@ impl<'a> AstDumper<'a> {
...
     }
 
     fn dump_expr_path(&mut self, expr: &ExprPathType) {
-        self.indent(|d| d.dump_expr(&expr.rhs));
         dump!(self, "path (::) @ {} {}", expr.pos, expr.id);
         self.indent(|d| d.dump_expr(&expr.lhs));
+        self.indent(|d| d.dump_expr(&expr.rhs));
```
dinfuehr, line 635: same here

dora/src/bytecode/generator.rs
```diff
@@ -693,10 +693,11 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
     ) -> Option<Register> {
         match *call_type {
             CallType::Method(_, _, _) => {
-                let obj_expr = expr.object().expect("method target required");
-                let reg = self.visit_expr(obj_expr, DataDest::Alloc);
-
-                Some(reg)
+                if expr.object().is_some() {
```
dinfuehr, line 696: Why did `CallType::Method` change, such that `e.object()` can be None now? I don't think this should be the case.

dinfuehr: For static methods this should be `CallType::Fct` instead.

dora/src/bytecode/generator.rs
```diff
@@ -805,7 +813,9 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
                     self.emit_invoke_direct(return_type, return_reg, fct_def_id, num_args);
                 } else if fct.is_virtual() {
                     self.emit_invoke_virtual(return_type, return_reg, fct_def_id, num_args);
-                } else if arg_bytecode_types[0] != BytecodeType::Ptr {
+                } else if arg_bytecode_types.is_empty()
```
dinfuehr, line 816: This change is probably also needed because of using `CallType::Method`.

dora-parser/src/ast/dump.rs
soc: I found it confusing that we were printing the rhs before the lhs ... I have no issue changing it back.
What was the reasoning doing it this way?

dora/src/bytecode/generator.rs
soc: I think these are the module methods – I'm using `Method` here already in preparation of modules becoming real values/instances that can extend classes and implement traits.

dinfuehr:
IMHO this PR would also be simpler and easier to review without the stdlib changes (still over 60 changed files). I guess the PR would break code without the stdlib changes, right? IIUC this could hopefully be solved in a backwards compatible way by allowing `IdentType::ClassAndModule` in check_expr_type_param of typeck/expr.rs and simply emitting IdentType::ClassType for it. This should allow adding this feature without touching exisiting library code. Tests can introduce their own modules with functions and don't need to use stuff from the stdlib. I think we should try to avoid depending on the stdlib too much since right now it could easily dramatically change at a later point and then we would lose test coverage for this feature.

You mentioned `@static` in traits as well: I agree this belongs in a separate PR but I think we should discuss how you plan to solve this here already.

dora/src/vm/src.rs
```diff
@@ -164,9 +165,11 @@ pub enum IdentType {
     // name of class with type params: SomeClass[T1, T2, ...]
     ClassType(ClassId, TypeList),
 
-    // name of method
+    // name of module
     Module(ModuleId),
 
+    ClassAndModule(ClassId, ModuleId),
```
dinfuehr, line 171: please add a short description here: e.g. "name of class and module"

dora-parser/src/ast/dump.rs
dinfuehr: No particular reason, I have an easier time understanding the tree if I pretend to rotate the screen to the right: then the lhs is left, the operator is in the middle and rhs is on the right. We can change this but then all uses need to change. Either way it doesn't belong in this PR.

dora/src/bytecode/generator.rs
dinfuehr: Does `ModuleName::foo()` then always pass the singleton object as an implicit argument? What about modules that are just used for namespace purposes?

Maybe it makes sense to add a new CallType, e.g. `CallType::ModuleMethod`?

dora-parser/src/ast/dump.rs
soc: Agreed, will remove it from this PR!

dora/src/bytecode/generator.rs
soc: It would only be used for non-namespace modules, as namespace modules cannot extend a class or implement traits – I think introducing a new type could make sense at this stage, even if it were removed it later on again.

dinfuehr: How do we distinguish namespace vs non-namespace modules?

I think `CallType::ModuleMethod` makes sense even with `self` as implicit argument just like in `CallType::Method`: The difference is that with `CallType::Method` there is an explicit object argument in the AST (`<obj>.method(<args>)`) whereas with module methods there isn't (`ModuleName::method(<args>)`).

soc:
@dinfuehr IMHO this PR would also be simpler and easier to review without the stdlib changes (still over 60 changed files). Would the PR break code without those changes? What's the reason?

Will do, felt like it made more sense with std lib adjusted, but you are right, keeping it our for now also makes sense. Will move it to another PR.

> You mentioned @static in traits as well: I agree this belongs in a separate PR but I think we should discuss how you plan to solve this here already.

The core idea is that after modules can implement traits, we can drop `@static` from traits themselves and instead have only traits with "normal" methods and impl them for the respective module. Instead e. g.

```
trait Default { @static fun default ... }
impl Default for Bool { @static fun default ... }
```
do
```
trait Default { fun default ... }
impl Default for module¹ Bool { fun default ... }
```

This approach avoids an inconsistency in which modules could impl traits without static members (and would turn them "static") and traits with static members (and keep them static), but classes could only impl the traits without static members (otherwise we are back to square one), and modules implementing traits with both static and non-static methods would muddy the water further and should be avoided too.

(This also suggests some improvements to bounds and static reasoning in the compiler, as it's quite useful to say things like "I want a T which has a module implementing trait X", which is not completely straight-forward due to inheritance.)

---

¹ likely not final syntax, I'm not quite happy with all the things I tried yet

dora/src/bytecode/generator.rs
soc: I think in general I think it's worth thinking whether it's desirable to have free-standing `fun`s in a namespace module vs. putting them in a "real" module, especially when looking at importing them ("static imports a la Java" have been the bane of my existence for quite a while).

Regarding the explicit object arg, I'd try to unify the syntax further in the future, because in the end `ModuleName::method(<args>)` doesn't really convey the info that, like other values `<obj>.method(<args>)`, it's an instance that the methods are called upon (just that `module`s have a "static" name that is known at compile-time).

Ideally I'd prefer having both use the same dot-syntax in the future `<obj>.method(<args>)` and `<module>.method(<args>)` to better convey this information. But that should surely be discussed elsewhere.

I'll introduce `ModuleMethod` and adjust the PR.

dora/src/cannon/codegen.rs
```diff
@@ -1601,10 +1601,7 @@ where
         assert_eq!(self.argument_stack.len() as u32, num);
         let arguments = self.argument_stack.drain(..).collect::<Vec<_>>();
         let self_register = arguments[0];
-
-        let bytecode_type_self = self.bytecode.register_type(self_register);
         let position = self.bytecode.offset_position(self.current_offset.to_u32());
-        assert_eq!(bytecode_type_self, BytecodeType::Ptr);
```
soc, line 1607: With this intact, it fails in `float0.dora` and `double0.dora` with:

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Int64`,
 right: `Ptr`', dora/src/cannon/codegen.rs:1615:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
RUN: target/debug/dora  --compiler=cannon tests/double0.dora
```

Offending method is `String::binary`.

dora/src/bytecode/generator.rs
soc: Yes, removed after adding `ModuleMethod`.

soc:
@dinfuehr Made the changes, now only module methods themselves and the bare minimum of changes to core types and the prelude is included to verify things work.

dinfuehr:
> @dinfuehr IMHO this PR would also be simpler and easier to review without the stdlib changes (still over 60 changed files). Would the PR break code without those changes? What's the reason?
> 
> Will do, felt like it made more sense with std lib adjusted, but you are right, keeping it our for now also makes sense. Will move it to another PR.
> 
> > You mentioned @static in traits as well: I agree this belongs in a separate PR but I think we should discuss how you plan to solve this here already.
> 
> The core idea is that after modules can implement traits, we can drop `@static` from traits themselves and instead have only traits with "normal" methods and impl them for the respective module. Instead of e. g.
> 
> ```
> trait Default { @static fun default ... }
> impl Default for Bool { @static fun default ... }
> ```
> 
> do
> 
> ```
> trait Default { fun default ... }
> impl Default for module¹ Bool { fun default ... }
> ```
> 
> This approach avoids an inconsistency in which modules could impl traits without static members (and would turn them "static") and traits with static members (and keep them static), but classes could only impl the traits without static members (otherwise we are back to square one), and modules implementing traits with both static and non-static methods would muddy the water further and should be avoided too.
> 
> So basically: _"there is only one kind of trait, with one kind of member (non-static), and traits can be implemented for classes and modules."_
> 
> (This also suggests some improvements to bounds and static reasoning in the compiler, as it's quite useful to say things like "I want a T which has a module implementing trait X", which is not completely straight-forward due to inheritance.)
> 
> ¹ likely not final syntax, I'm not quite happy with all the things I tried yet

Sounds good! I will take a look at the PR hopefully in the evening today...

soc:
@dinfuehr Great, thanks!

dora/src/cannon/codegen.rs
dinfuehr: InvokeDirect* checks whether the first parameter is nil and bails out if it is. The type of this argument therefore needs to be ptr. I think BytecodeGenerator emits the wrong bytecode: it should use InvokeStatic* if self isn't a ptr (see https://github.com/dinfuehr/dora/blob/f2557473b3d31944e21faf69606c92435fbd5577/dora/src/bytecode/generator.rs#L808-L811)

dora/src/bytecode/generator.rs
```diff
@@ -811,6 +811,22 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
+                let is_super_call = expr
+                    .object()
+                    .map(|object| object.is_super())
+                    .unwrap_or(false);
+
+                if is_super_call {
+                    self.emit_invoke_direct(return_type, return_reg, fct_def_id, num_args);
+                } else if fct.is_virtual() {
+                    self.emit_invoke_virtual(return_type, return_reg, fct_def_id, num_args);
+                } else if arg_bytecode_types.is_empty() {
```
dinfuehr, line 824: This probably needs to check whether the first argument is a pointer.

dora/src/bytecode/generator.rs
```diff
@@ -2292,6 +2316,10 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
                 specialize_type(self.vm, ty, &cls_type_params, type_params)
             }
 
+            CallType::ModuleMethod(_, _, ref type_params) => {
+                specialize_type(self.vm, ty, &TypeList::empty(), type_params)
```
dinfuehr, line 2320: I suppose this should look like Method. Could you please add a test for this?

dora/src/baseline/codegen.rs
```diff
@@ -1964,6 +1966,25 @@ where
                     }
                 }
 
+                CallType::ModuleMethod(_, fct_id, _) => {
+                    let fct = self.vm.fcts.idx(fct_id);
+                    let fct = fct.read();
+
+                    if fct.parent.is_trait() {
```
dinfuehr, line 1973: Is this even reachable at the moment?

dora/src/bytecode/generator.rs
```diff
@@ -811,6 +811,22 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
                     self.emit_invoke_direct(return_type, return_reg, fct_def_id, num_args);
                 }
             }
+            CallType::ModuleMethod(_, _, _) => {
+                let is_super_call = expr
```
dinfuehr, line 815: Do we really need to handle super and virtual calls here? What semantics does `ModuleName::methodName` have? IIUC this can't be super or virtual.

dora/src/bytecode/generator.rs
dinfuehr: Thinking about this: Right now there is no singleton object, so there is never self and therefore this should probably always emit self.

dora/src/bytecode/generator.rs
dinfuehr: If a module can't have type arguments right now, we could assert that the class does not have type parameters.

dora/src/semck/nameck.rs
```diff
@@ -359,7 +365,7 @@ mod tests {
         err(
             "fun Int() {}",
             pos(1, 1),
-            SemError::ShadowClassConstructor("Int".into()),
+            SemError::ShadowModule("Int".into()),
```
dinfuehr, line 368: Should we have a separate error for this: e.g. `ShadowClassAndModule`?

dora/src/typeck/expr.rs
```diff
@@ -1262,7 +1264,11 @@ impl<'a, 'ast> TypeCheck<'a, 'ast> {
                 CallType::Trait(trait_id, fct_id)
             } else {
                 let method_type = lookup.found_class_type().unwrap();
-                CallType::Method(method_type, fct_id, type_params.clone())
+                if method_type.is_module() {
+                    CallType::ModuleMethod(method_type, fct_id, type_params.clone())
```
dinfuehr, line 1268: We should have tests for module methods with type arguments as well. It would also be great to test a module for a generic class as well.

dinfuehr:
I've added a few comments. Could you also please remove the second commit (I suppose that's the commit that changes the stdlib), that should be trivial. I can't submit single commits from the PR.

dora/src/typeck/expr.rs
soc: I have an uncommitted `Assert` module that would work perfectly for this ...

dora/src/bytecode/generator.rs
soc: Yep ... it feels like a lot of these changes are living at an uneasy compromise between what would be sufficient/correct right now vs. what would be the desired behaviors in the future ...

dora/src/bytecode/generator.rs
soc: The idea is that modules can never have type arguments, at most their parent class can have them:

```
class Foo[T] { fun foo() -> T = ...
module Bar : Foo[Int] { fun foo() -> Int = ... ]
```

dora/src/baseline/codegen.rs
soc: Probably not, removed.

dora/src/bytecode/generator.rs
soc: Removed.

dora/src/semck/nameck.rs
soc: I think it is ok in this case, as the problem is really the module, not the class.

dinfuehr: Even when the class has a constructor?

dora/src/typeck/expr.rs
dinfuehr: I think I've already mentioned that we should try not to use stdlib functions in tests. Better to have a separate test specifically written for this.

dora/src/typeck/tests.rs
```diff
@@ -70,22 +70,36 @@ fn type_class_method_call() {
...
+            fun g() -> Int { return Foo::baz(); }");
+
+    err(
+        "module Foo {
+                 fun foo() -> Bar = Bar();
+               }
+               class Bar[T]
+
+               fun bar() {
+                 //Foo::foo(); // panicked at 'out-of-bounds', dora/src/ty.rs:732:33
```
soc, line 91: Stacktrace is
```
  13: <dora::ty::TypeList as core::ops::index::Index<usize>>::index
             at dora/src/ty.rs:732
  14: dora::semck::specialize::replace_type_param
             at dora/src/semck/specialize.rs:394
  15: dora::semck::specialize::replace_type_param::{{closure}}
             at dora/src/semck/specialize.rs:403
  16: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &mut F>::call_once
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/ops/function.rs:285
  17: core::option::Option<T>::map
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/option.rs:450
  18: <core::iter::adapters::Map<I,F> as core::iter::traits::iterator::Iterator>::next
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/iter/adapters/mod.rs:791
  19: <alloc::vec::Vec<T> as alloc::vec::SpecExtend<T,I>>::from_iter
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/liballoc/vec.rs:2007
  20: <alloc::vec::Vec<T> as core::iter::traits::collect::FromIterator<T>>::from_iter
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/liballoc/vec.rs:1919
  21: core::iter::traits::iterator::Iterator::collect
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/iter/traits/iterator.rs:1558
  22: dora::semck::specialize::replace_type_param
             at dora/src/semck/specialize.rs:401
  23: dora::typeck::lookup::MethodLookup::find
             at dora/src/typeck/lookup.rs:296
  24: dora::typeck::expr::TypeCheck::check_expr_call_method
             at dora/src/typeck/expr.rs:1259
  25: dora::typeck::expr::TypeCheck::check_expr_call
             at dora/src/typeck/expr.rs:990
  26: dora::typeck::expr::TypeCheck::check_expr
             at dora/src/typeck/expr.rs:2041
  27: <dora::typeck::expr::TypeCheck as dora_parser::ast::visit::Visitor>::visit_stmt
             at dora/src/typeck/expr.rs:2074
  28: dora::typeck::expr::TypeCheck::check
             at dora/src/typeck/expr.rs:41
  29: dora::typeck::check
             at dora/src/typeck.rs:36
  30: dora::semck::check
             at dora/src/semck.rs:111
  31: dora::test::parse_with_errors
             at dora/src/test.rs:44
  32: dora::semck::tests::err
             at dora/src/semck.rs:539
  33: dora::typeck::tests::type_module_method_call
             at dora/src/typeck/tests.rs:84
  34: dora::typeck::tests::type_module_method_call::{{closure}}
             at dora/src/typeck/tests.rs:73
```

dinfuehr: Hmm, the program should be invalid since `foo` returns `Bar`, however `Bar` should have a type argument. It seems this is not checked in modules however.

dora/src/typeck/tests.rs
```diff
@@ -70,22 +70,48 @@ fn type_class_method_call() {
...
+    );
+
+    ok("module Foo {
+                 fun foo[T]() -> Bar = Bar[T]();
+               }
+               class Bar[T]
+
+               fun bar() {
+                 // Foo::foo[Int](); // panicked at 'out-of-bounds', dora/src/ty.rs:732:33
+               }");
```
soc, line 107: Stacktrace is
```
  13: <dora::ty::TypeList as core::ops::index::Index<usize>>::index
             at dora/src/ty.rs:732
  14: dora::semck::specialize::replace_type_param
             at dora/src/semck/specialize.rs:394
  15: dora::semck::specialize::replace_type_param::{{closure}}
             at dora/src/semck/specialize.rs:403
  16: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &mut F>::call_once
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/ops/function.rs:285
  17: core::option::Option<T>::map
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/option.rs:450
  18: <core::iter::adapters::Map<I,F> as core::iter::traits::iterator::Iterator>::next
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/iter/adapters/mod.rs:791
  19: <alloc::vec::Vec<T> as alloc::vec::SpecExtend<T,I>>::from_iter
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/liballoc/vec.rs:2007
  20: <alloc::vec::Vec<T> as core::iter::traits::collect::FromIterator<T>>::from_iter
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/liballoc/vec.rs:1919
  21: core::iter::traits::iterator::Iterator::collect
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/iter/traits/iterator.rs:1558
  22: dora::semck::specialize::replace_type_param
             at dora/src/semck/specialize.rs:401
  23: dora::typeck::lookup::MethodLookup::find
             at dora/src/typeck/lookup.rs:296
  24: dora::typeck::expr::TypeCheck::check_expr_call_method
             at dora/src/typeck/expr.rs:1259
  25: dora::typeck::expr::TypeCheck::check_expr_call
             at dora/src/typeck/expr.rs:990
  26: dora::typeck::expr::TypeCheck::check_expr
             at dora/src/typeck/expr.rs:2041
  27: <dora::typeck::expr::TypeCheck as dora_parser::ast::visit::Visitor>::visit_stmt
             at dora/src/typeck/expr.rs:2074
  28: dora::typeck::expr::TypeCheck::check
             at dora/src/typeck/expr.rs:41
  29: dora::typeck::check
             at dora/src/typeck.rs:36
  30: dora::semck::check
             at dora/src/semck.rs:111
  31: dora::test::parse_with_errors
             at dora/src/test.rs:44
  32: dora::semck::tests::err
             at dora/src/semck.rs:539
  33: dora::typeck::tests::type_module_method_call
             at dora/src/typeck/tests.rs:84
  34: dora::typeck::tests::type_module_method_call::{{closure}}
             at dora/src/typeck/tests.rs:73
```

dora/src/typeck/tests.rs
soc: I updated the test, the `WrongNumberTypeParams` is reported correctly.

In the updated test, the generics are used correctly, but it still fails.

dora/src/typeck/tests.rs
soc: I updated the test, the WrongNumberTypeParams is reported correctly.

In the updated test, the generics are used correctly, but it still fails.

dora/src/typeck/tests.rs
soc: See below!

dora/src/typeck/tests.rs
soc: It fails with looking up type of `Bar`, at index 0.

dora/src/typeck/expr.rs
soc: Understood, I added a minimized test to typeck!

dora/src/typeck/lookup.rs
```diff
@@ -292,7 +292,7 @@ impl<'a, 'ast> MethodLookup<'a, 'ast> {
                 let list_id = self.vm.lists.lock().insert(cls_tps);
                 BuiltinType::Class(cls_id, list_id)
             }
-
+            LookupKind::Method(BuiltinType::Module(mid)) => BuiltinType::Module(mid),
```
soc, line 295: This would get me through type checking, but then it fails later in codegen:

```
  21: dora::semck::specialize::replace_type_param
             at dora/src/semck/specialize.rs:411
  22: dora::baseline::codegen::AstCodeGen::specialize_type
             at dora/src/baseline/codegen.rs:3474
  23: dora::baseline::codegen::AstCodeGen::determine_call_args_and_types
             at dora/src/baseline/codegen.rs:3617
  24: dora::baseline::codegen::AstCodeGen::build_call_site
             at dora/src/baseline/codegen.rs:3535
  25: dora::baseline::codegen::AstCodeGen::build_call_site_id
             at dora/src/baseline/codegen.rs:3563
  26: dora::baseline::codegen::AstCodeGen::emit_call
             at dora/src/baseline/codegen.rs:2026
  27: dora::baseline::codegen::AstCodeGen::emit_expr
             at dora/src/baseline/codegen.rs:718
  28: dora::baseline::codegen::AstCodeGen::emit_expr_result_reg
             at dora/src/baseline/codegen.rs:692
  29: dora::baseline::codegen::AstCodeGen::emit_stmt_expr
             at dora/src/baseline/codegen.rs:582
```

soc: Ok, this change is wrong anyway.

soc:
Another problem I found is that

```
module Foo {
  // note the missing [T] in the first occurrence of Bar:
  fun foo[T]() -> Bar = Bar[T]();
}
class Bar[T]
```

is not rejected ... though adding the `[T]` doesn't really help either.

soc:
@dinfuehr If you have time, could you have a look? It feels like I made some silly mistake and can't find it...

dinfuehr:
Yep, will take a look. Hopefully I will do this within the next few days...

soc:
@dinfuehr Thanks! I minimized the unrelated crash caused by a missing type parameter, it's https://github.com/dinfuehr/dora/issues/203.

soc:
Ok, rebased – this crashes the compiler on `baseline`, (`cannon` is fine):

```
module Foo {
  fun foo[T]() -> Bar[T] = Bar[T](23);
}

class Bar[T](let x: Int32)

fun main() {
  Foo::foo[Int32]();
}
```

Stacktrace:

```
  11: std::panicking::begin_panic
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libstd/panicking.rs:399
  12: <dora::ty::TypeList as core::ops::index::Index<usize>>::index
             at dora/src/ty.rs:722
  13: dora::semck::specialize::replace_type_param
             at dora/src/semck/specialize.rs:395
  14: dora::semck::specialize::replace_type_param::{{closure}}
             at dora/src/semck/specialize.rs:403
  15: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &mut F>::call_once
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/ops/function.rs:285
  16: core::option::Option<T>::map
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/option.rs:450
  17: <core::iter::adapters::Map<I,F> as core::iter::traits::iterator::Iterator>::next
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/iter/adapters/mod.rs:791
  18: <alloc::vec::Vec<T> as alloc::vec::SpecExtend<T,I>>::from_iter
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/liballoc/vec.rs:2007
  19: <alloc::vec::Vec<T> as core::iter::traits::collect::FromIterator<T>>::from_iter
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/liballoc/vec.rs:1919
  20: core::iter::traits::iterator::Iterator::collect
             at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7/src/libcore/iter/traits/iterator.rs:1558
  21: dora::semck::specialize::replace_type_param
             at dora/src/semck/specialize.rs:401
  22: dora::baseline::codegen::AstCodeGen::specialize_type
             at dora/src/baseline/codegen.rs:3468
  23: dora::baseline::codegen::AstCodeGen::determine_call_args_and_types
             at dora/src/baseline/codegen.rs:3611
  24: dora::baseline::codegen::AstCodeGen::build_call_site
             at dora/src/baseline/codegen.rs:3529
  25: dora::baseline::codegen::AstCodeGen::build_call_site_id
             at dora/src/baseline/codegen.rs:3557
  26: dora::baseline::codegen::AstCodeGen::emit_call
             at dora/src/baseline/codegen.rs:2019
  27: dora::baseline::codegen::AstCodeGen::emit_expr
             at dora/src/baseline/codegen.rs:718
  28: dora::baseline::codegen::AstCodeGen::emit_expr_result_reg
             at dora/src/baseline/codegen.rs:692
  29: dora::baseline::codegen::AstCodeGen::emit_stmt_expr
```

soc:
Adding some print-statements to replace_type_param:

```
        BuiltinType::FctTypeParam(fct_id, tpid) => {
            let f = vm.fcts.idx(fct_id);
            let n =  vm.interner.str(f.read().name);
            if n.as_str() == "foo" {
                println!("{}, tp_list_len: {}, idx: {}", n, fct_tp.len(), tpid.idx());
            }
            fct_tp[tpid.idx()]
        },
```

shows that the cannon compiler runs through this part of the code 8 times, and the indices are correct:

```
RUST_BACKTRACE=1 target/debug/dora tests/module/module2.dora
foo, tp_list_len: 1, idx: 0
foo, tp_list_len: 1, idx: 0
foo, tp_list_len: 1, idx: 0
foo, tp_list_len: 1, idx: 0
foo, tp_list_len: 1, idx: 0
foo, tp_list_len: 1, idx: 0
foo, tp_list_len: 1, idx: 0
foo, tp_list_len: 1, idx: 0
```

But with baseline compiler, it fails at the second:

```
RUST_BACKTRACE=1 target/debug/dora --compiler=baseline tests/module/module2.dora                                                                                                 
foo, tp_list_len: 1, idx: 0
foo, tp_list_len: 0, idx: 0
thread 'main' panicked at 'type list index out-of-bounds', dora/src/ty.rs:722:33
```

soc:
I rebased and fixed the remaining issue (`specialize_for_call_type` was a no-op for `ModuleMethod`s).

dora/src/cannon/codegen.rs
```diff
@@ -1978,10 +1978,7 @@ where
 
         let arguments = self.argument_stack.drain(..).collect::<Vec<_>>();
         let self_register = arguments[0];
-
-        let bytecode_type_self = self.bytecode.register_type(self_register);
         let position = self.bytecode.offset_position(self.current_offset.to_u32());
-        assert_eq!(bytecode_type_self, BytecodeType::Ptr);
```
dinfuehr, line 1984: This still needs to be solved. The first argument needs to be a pointer.

dora/src/typeck/tests.rs
dinfuehr: IIRC this was fixed, right?

tests/module/module2.dora
```diff
@@ -0,0 +1,11 @@
+//= cannon
```
dinfuehr, line 1: FYI: The line `//= cannon` isn't necessary anymore, all tests are now run with both compilers. No strong need to remove this line here (although you can if you want). I plan to remove those lines in all tests anyway.

dora/src/cannon/codegen.rs
soc: Ok, will look into it, thanks!

tests/module/module2.dora
soc: Ok, thanks!

dora/src/typeck/tests.rs
soc: Yes, I'll resolved this.

dora/src/cannon/codegen.rs
soc: Was some left-over thing, removed it and everything is fine now!

soc:
@dinfuehr Resolved the points you mentioned!

dinfuehr:
Thanks!

PR #194, opened 2020-03-29
Parse packages and imports

soc:


dinfuehr:
I want to use the name package for libraries or translation units. The way I envision this there would be no package statement and packages can't be nested but they can depend on other packages. I think two concepts should be enough here: packages and modules.

In this PR it looks like packages would be more Java-like, modules are a separate thing and we would need a third concept for a library/translation unit then.

soc:
> The way I envision this there would be no package statement and packages can't be nested but they can depend on other packages.

This is an interesting idea!

How would code be organized in cases where you would traditionally nest things? E. g. when you have a `time` package and formatting time is in `time.format`, and handling timezones is in `time.zone`, "traditionally" it would probably look like this:

```
time/
    LocalTime.dora: // package time; class LocalTime(...)
    format/
        DateTimeFormatter.dora: // package time.format; module DateTimeFormatter { ... }
    zone/
        ZonedDateTime.dora: // package time.zone; class ZonedDateTime(...)
```

How would this look like with your approach?

> In this PR it looks like packages would be more Java-like, modules are a separate thing and we would need a third concept for a library/translation unit then.

I could rename `package` to `module` form the get-go instead of later, such that we only have package-as-library and modules.

---

We still need `import`s, even in your design, right?

dinfuehr:
I think the solution is quite traditional. `time` would be the package, `format` and `zone` would be modules then. Modules could still be nested. Just to be clear: If `format` or `zone` need to be in a different package than `time`, then they can't use the path `time::format`.

I think in Dora we could already have three packages internally: the main app, standard library and boots. This means there could already be support for packages without exposing anything. Each package should be "type checkable" on its own (+its dependencies).

There would still be imports. I guess we could replace `package` with `module` here, but then I am not sure yet how we would best include files.

dinfuehr:
Could you reduce this PR to the commit that actually deals with packages? It shouldn't depend on any of the other work.

soc:
> I think the solution is quite traditional. time would be the package, format and zone would be modules then. Modules could still be nested. Just to be clear: If format or zone need to be in a different package than time, then they can't use the path time::format.

Ah ok. Understood!

Then it's really just a matter of renaming `package` to `module` to get out of your way for your use  of `package`-as-library.


> I guess we could replace package with module here, but then I am not sure yet how we would best include files.

Mhh, what do you mean with including files? In which case would an import not be sufficient?

soc:
I'm reworking this – I'll ping again when this is ready.

soc:
After some further investigation, I think that the complexity of sorting out whether `module` is a namespace declaration or a "real" module isn't worth it – both for the implementation and for users.

I'll still add `import`, but the plan for namespacing is now to solely rely on the directory structure and not use any in-source declaration at all. This has the additional benefit that unlike with an in-source declaration, the dir structure and the decl can't go out of sync.

I'll overhaul this PR accordingly.

soc:
This has some implications for copy-paste-ability, but I think it's worth it.

Additionally, the harm is substantially reduced by not "automagically" importing things from the current context (as it is done by many other languages).

soc:
@dinfuehr I see you went ahead with the original plan I discarded in https://github.com/dinfuehr/dora/commit/410ea8eee0df5ef8b5f754adca816e31a2754fba – do you disagree on the verdict or something else?

dinfuehr:
The concrete syntax doesn't matter much. The `namespace` syntax that I implemented is easy enough for now and allows for simple unit testing. I would also like to use this soon in boots. Once namespace works, switching this over to another syntax should be relatively easy. That said, right now I would say that we should allow for in-source declarations as well.

soc:
Ah ok.

dinfuehr:
We already have a method for loading all dora files in a directory in the codebase. However I would also like to add unit tests in typeck/tests.rs and other files in semck. In-source declaration makes it easy to write those tests in a concise way.

I would expect most of the work to be orthogonal to that decision anyway: We need a symbol table for the namespace, need to allow paths (so `a::b::c`) instead of simple identifiers in many places, support for private/pub elements in the namespace, importing stuff from other namespaces, etc. This will be the bulk of the work.

soc:
> We already have a method for loading all dora files in a directory in the codebase.

Oh, where? I didn't see that.

> [...] This will be the bulk of the work.

Agreed!

soc:
Let's close this, this is completely outdated.

PR #196, opened 2020-04-11
Add tests: using blocks for field initialization

soc:
This brings syntax for fields (`let`, `var`), closer to syntax for `fun`s.

### DIFF
--- a/tests/field-block.dora
+++ b/tests/field-block.dora
@@ -1,0 +1,15 @@
+//= cannon
+
+fun main() {
+  let a = A::empty();
+
+  assert(a.lx == 1);
+  assert(a.vx == 2);
+}
+
+class A(le: Int, va: Int) {
+  let lx: Int = { le + 1 };
+  var vx: Int = { va + 1 };
+
+  @static fun empty() -> A = A(0, 1);
+}


soc:

dora-parser/src/parser.rs
```diff
@@ -1130,6 +1130,10 @@ impl<'a> Parser<'a> {
             self.expect_token(TokenKind::Eq)?;
             let expr = self.parse_expression()?;
 
+            Ok(Some(expr))
+        } else if self.token.is(TokenKind::LBrace) {
```
dinfuehr, line 1134: Does this even need special handling? The block is already an expression and should be handles in the `self.parse_expression()` call above?

soc: Ugh ... yes, you are absolutely right. I'll remove everything but the new tests. Sorry!

soc:
@dinfuehr done!

tests/field-block.dora
```diff
@@ -0,0 +1,21 @@
...
+  assert(a.lx == 1);
+  assert(a.ly == 2);
+
+  assert(a.vx == 1);
+  assert(a.vy == 2);
+}
+
+class A(le: Int, va: Int) {
+  let lx: Int = { le + 1 };
+  let ly: Int = { self.lx + 1 };
```
dinfuehr, line 15: Not sure about this: For removing `nil`, I would like to remove access to `self` in the initializer expressions. This way we could easily ensure that there is no access to uninitialized fields (we can probably relax this restriction later). This test would introduce more usages of `self` and would make that harder.

soc: @dinfuehr done!

PR #197, opened 2020-04-11
test-runner: allow to define timeout for test cases

hzahradnik:
This PR adds the possibility to define a time-out for a test file  with `//= timeout <value>`. Currently, the timeout is optional. Tests without it, run as before. Should we set a time-out as default, so we can guarantee, that the tests don't run indefinitely?

The function to start the process within a time-out block is from https://gist.github.com/pasela/9392115. I used this, as there is a lot of text to find on problems with time-out in ruby (e.g. https://redmine.ruby-lang.org/issues/4681)  and it would be pointless to invest a lot of time in this problem. ;)

### DIFF
--- a/tools/tester.rb
+++ b/tools/tester.rb
@@ -41,30 +41,28 @@ class TestUtility
     out_reader = nil
     err_reader = nil
 
-    begin
-      Open3.popen3(cmd) do | stdin, stdout, stderr, wait_thr |
-        Timeout.timeout(timeout) do
-          result[:pid] = wait_thr.pid
+    Open3.popen3(cmd) do | stdin, stdout, stderr, wait_thr |
+      Timeout.timeout(timeout) do
+        result[:pid] = wait_thr.pid
 
-          stdin.close
-          out_reader = Thread.new { stdout.read }
-          err_reader = Thread.new { stderr.read }
+        stdin.close
+        out_reader = Thread.new { stdout.read }
+        err_reader = Thread.new { stderr.read }
 
-          result[:status] = wait_thr.value
-        end
-      rescue Timeout::Error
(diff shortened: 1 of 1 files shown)


tools/tester.rb
```diff
@@ -103,14 +213,24 @@ def print_results()
   def run_test(optional_vm_args, mutex)
     temp_out = Tempfile.new("dora-test-runner")
     cmdline = "#{binary} #{vm_args} #{optional_vm_args} #{test_file} #{args}"
-    stdout, stderr, status = Open3.capture3(cmdline)
-    result = check_test_run_result(stdout, stderr, status)
+    if timeout != nil
```
dinfuehr, line 216: Let's enforce a default timeout for every test (e.g. 60 seconds - not sure whether this makes sense).

hzahradnik: I set the default to 60 seconds. Hopefully, this time the actions do not fail.

tools/tester.rb
```diff
@@ -248,6 +298,16 @@ def run_tests
 
   ret = failed == 0
 
+
+  if faillist.any?
+    puts "failed tests:"
+
+    faillist.each { |test_case| test_case.results
```
dinfuehr, line 305: This is pretty hard to understand for me. AFAIR Ruby convention is to use begin/end instead of {} when the block spans multiple lines. I would suggest to write this something like this:

```ruby
for test_case in faillist
  for run_name, run_result in test_case.results
    next if run_result == true
    puts "  #{test_case.file}.#{run_name}"
  end
end
```

dinfuehr:
Thank you very much 👍 I think this PR will be incredibly valuable.

PR #198, opened 2020-04-12
Cleanup parser

soc:
- Try to sort the keywords into categories.
- Made `parse_comma_list` a bit more generic by allowing to specify the separator.
- Drop `~` which is not used anywhere, as we use `!` as bitwise negation.
- Drop `+=`, as it's not used – given how it makes it harder to read code (another precedence rule to keep in mind) I also think it would be better to simply not have it (just like we do not have `++`, `--`).

### DIFF
--- a/dora-parser/src/ast.rs
+++ b/dora-parser/src/ast.rs
@@ -662,7 +662,6 @@ pub enum Stmt {
     StmtBreak(StmtBreakType),
     StmtContinue(StmtContinueType),
     StmtReturn(StmtReturnType),
-    StmtDefer(StmtDeferType),
     StmtFor(StmtForType),
 }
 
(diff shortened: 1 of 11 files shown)


soc:
@dinfuehr What is your position on `defer`?

If I have understood the purpose of `defer` correctly (inferred from looking at other languages that have it), I think it would be reasonable to drop the keyword:

 I believe it's nicer to deal with this with more general language features (lambdas); i. e. instead of

```
let conn = socket.openConnection(...);
defer conn.close();
// do things with conn
```

write

```
let conn = socket.connection(...);
conn.run(|openConnection| => /* do things with conn */)
```
with

```
class Socket {
  fun run(op: Connection => ...) {
    let openConnection = self.open();
    op(openConnection);
    let openConnection.close();
  }
}
```

As we don't support catching exceptions, one of the main issue other languages have with such a construct (people can "circumvent" the `close()` by throwing an exception from the `op` lambda) does not apply to us.

As exceptions are fatal in Dora, we don't have this danger of having the program keep running with violated post-conditions.

soc:
@dinfuehr I added another commit dropping – I'll happily remove it from this PR if you disagree or believe further discussion is required.

dinfuehr:
Agreed, we can get rid of `defer` at least for now, especially since it's not working atm. It might still makes sense to have it in the future, but let's revisit this later and only when we see there is a clear need for it.

soc:
Great, thanks!

dinfuehr:
Thanks!

PR #199, opened 2020-04-17
cannon: refactor store and load register to method

hzahradnik:
As discussed, I will split #195 into several smaller PR to reduce the complexity. The first PR refactors the store and load to the bytecode registers to a explicit function.


dinfuehr:
Very cool!

PR #200, opened 2020-04-18
bytecode: PushRegister to use register as function argument

hzahradnik:
Currently, the bytecode requires the registers for function arguments to be a list of consecutive registers. As it is not guaranteed, that the arguments are in order, a register chain was allocated and the values were copied. With PushRegister the used registers are explicitly marked instead of copied to a new register.


dora/src/cannon/codegen.rs
```diff
@@ -1575,38 +1574,35 @@ where
...
         fct_def_id: FctDefId,
-        start_reg: Register,
         num: u32,
         bytecode_type: Option<BytecodeType>,
     ) -> AnyReg {
         assert!(num > 0);
 
-        let bytecode_type_self = self.bytecode.register_type(start_reg);
+        assert_eq!(self.argument_stack.len() as u32, num);
+        let arguments = self.argument_stack.drain(..).collect::<Vec<_>>();
```
dinfuehr, line 1598: I think you can take ownership of the vector (and replace the current one with an empty one) with: `let arguments = mem::replace(&mut self.argument_stack, Vec::new())`.

dora/src/cannon/codegen.rs
```diff
@@ -1683,10 +1672,13 @@ where
         let fct = fct.read();
 
         if let FctKind::Builtin(intrinsic) = fct.kind {
-            return self.emit_invoke_intrinsic(&*fct, &*fct_def, intrinsic, start_reg, num);
+            return self.emit_invoke_intrinsic(&*fct, &*fct_def, intrinsic, num);
         }
 
-        let argsize = self.emit_invoke_arguments(start_reg, num);
+        assert_eq!(self.argument_stack.len() as u32, num);
+        let arguments = self.argument_stack.drain(..).collect::<Vec<_>>();
```
dinfuehr, line 1679: Here as well

dora-boots/bytecode_dump.dora
```diff
@@ -86,14 +86,14 @@ class BytecodeDumper(let bc: BytecodeFunction): BytecodeVisitor {
         println(" ${r1}, ${gid.value}");
     }
 
-    fun emitFctVoid(name: String, fid: FctId, r1: BytecodeRegister, cnt: Int) {
+    fun emitFctVoid(name: String, fid: FctId, cnt: Int) {
```
dinfuehr, line 89: I think it's time to also get rid of the `count` argument, not in this PR though.

dinfuehr:
Very cool! I've already merged it, fix the comments in a subsequent PR.

PR #201, opened 2020-04-19
bytecode: remove argument count from Invoke*

hzahradnik:
After removing the register chain for arguments and explicitly pushing the argument registers, it is possible to remove the count argument for all Invoke bytecodes.

To ensure we have the correct number of arguments pushed, I added the parameter types to the specialized function (`FctDef`) and then compare the number of non-unit types with the number of pushed arguments. Right now, I didn't refactor the code in the bytecode generator to use this `Vec<BuiltinTypes>`, but theoretically this should allow to use the specialized types without specializing everywhere we need them. 
What do you think about this change? Maybe it could make sense to also add the specialized return type?

### DIFF
--- a/dora-boots/bytecode_dump.dora
+++ b/dora-boots/bytecode_dump.dora
@@ -86,14 +86,14 @@ class BytecodeDumper(let bc: BytecodeFunction): BytecodeVisitor {
         println(" ${r1}, ${gid.value}");
     }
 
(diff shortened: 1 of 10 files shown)


dinfuehr:
I think it would be good to assert this in debug mode but I wouldn't store it in FctDef. At some point we also need to care about memory usage and the concrete parameters should be simple/fast enough to recompute.

With bytecode verification we would need to verify not only the argument count but also the types.

PR #202, opened 2020-05-09
Drop `var` keyword in parameter list

soc:
As discussed earlier!

### DIFF
--- a/bench/binarytrees/binarytrees.dora
+++ b/bench/binarytrees/binarytrees.dora
@@ -44,7 +44,8 @@ fun loops(iterations: Int, depth: Int) {
     depth.toString() + "\t check: " + check.toString());
 }
 
-fun createTree(var depth: Int) -> TreeNode {
+fun createTree(depth: Int) -> TreeNode {
+  var depth = depth;
   let node = TreeNode(0, nil, nil);
 
   if depth > 0 {
--- a/bench/gcbench/gcbench.dora
+++ b/bench/gcbench/gcbench.dora
@@ -21,15 +21,16 @@ fun numIters(i: Int) -> Int {
     return 4 * treeSize(stretchTreeDepth) / treeSize(i);
 }
 
-fun populate(var iDepth: Int, thisNode: Node) {
-    if iDepth <= 0 {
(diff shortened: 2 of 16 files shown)


dora/src/error/msg.rs
```diff
@@ -485,12 +485,16 @@ impl SemErrorAndPos {
...
 
     pub fn message(&self, vm: &VM) -> String {
-        let file = vm.file(self.file);
-        format!(
-            "error in {} at {}: {}",
-            file.name,
-            self.pos,
-            self.msg.message()
-        )
+        let file = self.filename(vm, self.file);
```
dinfuehr, line 488: Thanks for tackling this! It was indeed very annoying. However I don't think this is the right way to fix this. I have a commit here that should fix this problem: https://github.com/dinfuehr/dora/commit/3e7ccbf364b3c7514261387595ec2c00104a3931

bench/binarytrees/binarytrees.dora
```diff
@@ -44,7 +44,8 @@ fun loops(iterations: Int, depth: Int) {
     depth.toString() + "\t check: " + check.toString());
 }
 
-fun createTree(var depth: Int) -> TreeNode {
+fun createTree(dep: Int) -> TreeNode {
+  var depth = dep;
```
dinfuehr, line 48: For all those replacements I would prefer if we would just have: `var depth = depth`, so not renaming the variable.

dora/src/error/msg.rs
soc: Great, thanks, will drop this!

bench/binarytrees/binarytrees.dora
soc: Ah, right, completely forgot we can shadow like this. That's of course much better, will do it!

soc: Done!

soc:
@dinfuehr Oh, the conflicts are new. :-( Will fix them.

dinfuehr:
Thanks for pinging me! Looks good, thanks!

soc:
Thanks!

PR #204, opened 2020-05-31
benchmark: implemented richards from octane benchmarks

hzahradnik:


bench/richards/richards.dora
```diff
@@ -0,0 +1,347 @@
...
+        scheduler.schedule();
+
+        if scheduler.queueCount != EXPECTED_QUEUE_COUNT || scheduler.holdCount != EXPECTED_HOLD_COUNT {
+            println("Error during execution: queueCount = ${scheduler.queueCount}, holdCount = ${scheduler.holdCount}.");
+            exit(1);
+        }
+    }
+}
+
+var ID_IDLE: Int32 = 0;
```
dinfuehr, line 52: those could be constants

bench/richards/richards.dora
```diff
@@ -0,0 +1,347 @@
...
+var NUMBER_OF_IDS: Int32 = 6;
+
+var KIND_DEVICE: Int32 = 0;
+var KIND_WORK: Int32 = 1;
+
+class Scheduler() {
+    var queueCount: Int32 = 0;
+    var holdCount: Int32 = 0;
+    var blocks: Array[Option[TaskControlBlock]] = arrayFill[Option[TaskControlBlock]](NUMBER_OF_IDS.toInt64(), none[TaskControlBlock]());
+    var currentTcb: Option[TaskControlBlock] = none[TaskControlBlock]();
```
dinfuehr, line 67: Option is still quite expensive in Dora relative to using `nil`, so I don't expect the Dora version to be competitive with that of JS. But we can keep the code like this. IIRC for binarytrees I measured a slowdown of factor 5, it's probably less here though.

bench/richards/richards.dora
hzahradnik: Done.

bench/richards/richards.dora
hzahradnik: I recall you mentioning the expensive unwrapping of `Option`, but decided to use it anyway, as `nil` could possibly be removed in the long run.

hzahradnik:
As asked by you, the number of iterations is now a possible command line argument. Additionally, I added a test case, which runs the benchmark with 10 iterations.

dinfuehr:
Thanks for adding the test! I was asking for this since each benchmark not run as part of the test suite got broken quite fast...

PR #205, opened 2020-06-02
Implement intrinsic for total order comparison on floating point types.

soc:


soc:
Looks like I'm an idiot. :-)

```
            Intrinsic::FloatSrt | Intrinsic::DoubleSrt => {
                let mode = if intr == Intrinsic::DoubleCmp {
```

That's ... not going to work.

I'm fixing this and giving it another try!

soc:
Welp, looks like it's broken on cannon now – I can't sort out what the problem here is...

soc:
Looks like our current implementation is buggy too (order of NaNs is incorrect), see comment in added test in https://github.com/dinfuehr/dora/pull/255.

soc:
Welp. As mentioned on Gitter, the use of legacy instructions that clobbered a fixed register was the reason the code didn't work for ages.

soc:

soc:
As requested, I replaced the macro assembler calls with direct assembler calls, could you have a look, @dinfuehr?

soc:
@dinfuehr ping?

soc:
@dinfuehr ping?

soc:
CI fail is due to Ruby install: `Error: Version 2.6.x not found`

dinfuehr:
Looks like this one needs to be updated. Then we can merge it.

soc:
@dinfuehr Rebased.

soc:
@dinfuehr I rebased the commit on master such that the fixed CI could run again, can you have a look?

dinfuehr:
Hey, had to revert this PR because the tests didn't work on arm64.

soc:
What was the error?

dinfuehr:
I think it was assertion failures in the sorting tests.

soc:
I think that's the end of this journey, fyi I closed my issues that I'd class as suggestion, and unsubscribed from those I'd consider bug reports.

dinfuehr:
Nevertheless thanks!

soc:
The experience as a contributor was simply a magnitude worse than whatever I had seen before. Wish you the best, and hope you can figure things out.

It's sad, because I thought this language had potential, but in the end I have other projects where I feel my efforts are actually appreciated and not treated as a nuisance.

Let me know if you want dora-lang.org then I transfer it, otherwise I'll let the registration run out.

PR #206, opened 2020-06-06
Module functions – migrate @static methods in standard library 

soc:


dinfuehr:
Let's merge that first commit first and add some tests with it as well.

soc:
@dinfuehr Added tests. Remaining commits are in https://github.com/dinfuehr/dora/pull/207.

PR #207, opened 2020-06-08
Module functions – migrate @static methods in standard library

soc:

PR #208, opened 2020-06-08
Implement Assert

soc:


soc:
Addition after https://github.com/dinfuehr/dora/pull/206 and https://github.com/dinfuehr/dora/pull/207.

PR #213, opened 2020-06-11
Unified Condition Expressions, part 1

soc:
This is work toward https://github.com/dinfuehr/dora/issues/269.

- Allow multiple if-cases.
- Conditions can be split into a common discriminator and continuations in individual if-cases.
  `...` is used to indicate this, but this is provisionally and could be changed to something else (including no token at all).


soc:
@dinfuehr ping!

dora/src/language/generator.rs
```diff
@@ -871,40 +879,60 @@ impl<'a> AstBytecodeGen<'a> {
...
             }
 
-            self.gen.bind_label(else_lbl);
             self.visit_expr(else_block, DataDest::Reg(dest));
             self.gen.bind_label(end_lbl);
 
             dest
         } else {
-            // Without else-branch there can't be return value
+            // Without else-branch there can't be return value --- not true anymore, if cond_head is an enum, branches can be exhaustive
```
soc, line 897: The comment is probably a bit premature, as support for enums hasn't been implemented yet.

dora/src/language/generator.rs
```diff
@@ -871,40 +879,60 @@ impl<'a> AstBytecodeGen<'a> {
...
             assert!(ty.is_unit());
 
             let end_lbl = self.gen.create_label();
-            let cond_reg = self.visit_expr(&expr.cond, DataDest::Alloc);
+            let cond_reg = self.visit_expr(&expr.cond_head.expr.as_ref().unwrap(), DataDest::Alloc);
             self.gen.emit_jump_if_false(cond_reg, end_lbl);
             self.free_if_temp(cond_reg);
 
-            self.emit_expr_for_effect(&expr.then_block);
+            self.emit_expr_for_effect(&expr.branches.get(0).unwrap().then_block); // fixme
```
soc, line 905: Best way to address this for now?

tests/if/if7.dora
```diff
@@ -0,0 +1,14 @@
+fun main() {
+  let x = if "abc"
+    ... == "bcd" { 1 }
+    ... != "abc" { 2 }
+    ... .size() == 3L { 3 }
+    else { -1 };
+  assert(x == 3);
+  let x = if "abc"
```
soc, line 8: I should add an test where the cond_head is a function call to show that the hoisting into a let works as expected and the function is only executed once.

soc:
@dinfuehr ping?

dinfuehr:
Not sure what to do with this PR. It doesn't look fully fledged out, which is kinda problematic since if expressions are so fundamental.

soc:
@dinfuehr I'm waiting for your thoughts on the two sections I marked.

tests/if/if7.dora
soc: Done.

dora/src/language/generator.rs
soc: Reworded.

dora/src/language/generator.rs
soc: Changed.

soc:
@dinfuehr I rebased the commit on master such that the fixed CI could run again, can you have a look?

PR #219, opened 2020-06-13
Fixed duplication in dependencies.

ashwanidausodia:
When you run `cargo tree -d` (requires cargo 1.44+) , you see a lot of
duplication is dependencies. This duplication usually affects both
compile times and the binary size. Here, the major reason for
duplication was older version of `docopt` in dora and `parking_lot` in
dora-parser. Also, this older version of `docopt` uses `rustc_serialize`
which is deprecated in favour of `serde`.
https://users.rust-lang.org/t/deprecation-of-rustc-serialize/10485?u=dtolnay

I have upgraded the `docopt`, `parking_lot` and patch versions for some
other dependencies. There are no duplications now. It reduces the total
number of dependencies from 106 to 87.

Although there was negligible improvement in build time, it reduces the
size of release binary by more than 1MB (for `amd64-linux` target).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,27 +2,12 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.6.10"
+version = "0.7.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


dinfuehr:
Great work, thanks! Even if build time is negligible on x64, it might help more on my Raspberry Pi.

PR #220, opened 2020-06-16
Return Option for HashMap#remove and Bool for HashSet#remove

soc:
In case of HashMap, this further reduces the amount of `nil`s we produce.


dinfuehr:
Cool, thanks!

PR #221, opened 2020-06-16
Introduce Sizable trait, make existing classes use it where possible¹

soc:
The decision between length() and size() was based on:

- Stringable sounds a bit weird, Lengthable sounds very weird
- size() makes more sense on more types than length() does
- size() is a little bit shorter than length()

¹ We currently cannot define traits for generic classes –
  the impls for them have been provided, but commented out.

### DIFF
--- a/bench/falsesharing/falsesharing.dora
+++ b/bench/falsesharing/falsesharing.dora
@@ -22,14 +22,14 @@ fun construct(size: Int) -> Array<Foo> {
 
     var i = 0;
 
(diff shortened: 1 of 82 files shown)


dinfuehr:
After merging 220 there is a merge conflict here, I tried to fix unsuccessfully. Should be pretty easy though, can you please rebase this PR?

soc:
Will do!

soc:
@dinfuehr Rebased.

PR #222, opened 2020-06-22
Allow constructor varargs and use them

soc:


dinfuehr:
My first reaction after looking at the failed tests was that it's probably safer for now to not break already existing tests (even with the old baseline compiler). If this is too much work, we could consider dropping baseline entirely though. Agreed, implementing the array constructor as an intrinsic should simplify the code and make it easier to change in the future.

soc:
Will do!

soc:
@dinfuehr Dropping baseline is https://github.com/dinfuehr/dora/pull/223.

I also removed the third commit with the array stuff, so that this PR with the first two commits should be fine know.

soc:
@dinfuehr Rebased, could you have a look?

dinfuehr:
Thanks, looks good! Could you please test the case where the varargs argument isn't the last argument in a constructor. There is one for regular functions already.

soc:
Will do!

PR #223, opened 2020-06-24
Drop baseline compiler

soc:


dinfuehr:
Dropping baseline was meant as a last resort, we didn't even try to update baseline as well. Let's keep it for now.

dinfuehr:
Thank you very much 👍

soc:
No problem! Did you find the differences between baseline and cannon you wanted to sort out?

dinfuehr:
Not in too much detail TBH. But I think with our limited time it is not worth to keep the second compiler around for much longer

PR #225, opened 2020-07-31
 Migrate Array(<len>) constructor calls to Array::ofSizeUnsafe(<len>) method calls

soc:


dinfuehr:
Thanks!

soc:
Great, thanks!

PR #226, opened 2020-08-01
Topic/cstr varargs array part2

soc:


dora/src/bytecode/generator.rs
```diff
@@ -1548,6 +1549,13 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
 
                 Intrinsic::GenericArrayNew => self.emit_intrinsic_new_array(expr, dest),
 
+                Intrinsic::GenericArrayWithValues => {
+                    let ty = self.ty(expr.id);
+                    let array_reg = self.emit_array_with_variadic_arguments(expr, &[ty], 0, dest);
+                    self.registers.free_if_temp(array_reg);
```
dinfuehr, line 1555: I don't think this line is correct.

dora/src/bytecode/generator.rs
```diff
@@ -1580,24 +1588,11 @@ impl<'a, 'ast> AstBytecodeGen<'a, 'ast> {
...
-        let callee = self.vm.fcts.idx(callee_id);
-        let callee = callee.read();
-
-        // Determine types for arguments and return values
-        let (arg_types, _, return_type) = self.determine_callee_types(&call_type, &*callee);
-
-        assert!(return_type.is_unit());
-        assert_eq!(arg_types.len(), 2);
+        // We need array of elements
+        let element_ty = self.ty(expr.id);
```
dinfuehr, line 1592: Nice!

dora/src/bytecode/generator.rs
soc: So this line needs to happen after the caller got the register (and likely did something with it), right?

dinfuehr: I think you can simply remove that line. array_reg can't be freed at that point, it should be freed by the caller later. Unfortunately there is no check in the bytecode generator yet that makes sure that instruction operands are still live and not freed.

dinfuehr: Btw the caller should already free the register, so you shouldn't need to change any other code.

soc: I think without that line I ran into the assert – I'll check again.

soc: Weeeeird – it works! I would have sworn that didn't work earlier ...

dinfuehr: That was without the dest argument, right?

soc:
@dinfuehr This should be fine now, right?

Could you merge https://github.com/dinfuehr/dora/pull/227 before this one, so I can rebase it cleanly?

soc:
Oops, yes, sorry. Will fix.

soc:
@dinfuehr Fixed!

PR #228, opened 2020-08-13
Fix compilation errors in falsesharing.dora

soc:
(still asserts at runtime)


soc:
(thought this was better than nothing... didn't want to spend too much time on it, as it doesn't seem to run with the test suite anyway)

dinfuehr:
Thanks!

PR #230, opened 2020-09-20
Merge KnownClasses/KnownModules/KnownFunctions into KnownElements

soc:
This appears to have been some historical leftover – that is now removed.


dinfuehr:
Not sure about this, since it was supposed to go into the other direction: Split one struct into multiple different ones.

soc:
Oh ok, so shall I move it the other way around then?

dinfuehr:
You can of course - if you want to.

PR #232, opened 2020-09-26
Switch from -> to : for result types

soc:


soc:

dinfuehr:
Thanks!

PR #233, opened 2020-09-26
Switch from : to extends for parent classes

soc:


soc:
(based on the `->`-to-`:` PR, so the first 8 commits can be ignored)

soc:
@dinfuehr Rebased!

soc:
Thanks!

PR #234, opened 2020-09-27
Parse direct trait impls in classes and modules

soc:


soc:
Currently fails for

```
class Foo[T]
impl MyTrait for ...
```

I think some temporary adjustment to require classes/modules/... to either have a `{...}` or a `;` could make sense at this stage.
(Mid-to-long-term I want to work on semicolon inference, so then this will come up anyway.)

dinfuehr:
We could simply have temporary adjustments for this feature instead. Then no current code would need to change. Not sure that automatic semicolon insertion would help here.

We could fix this by using `implements` instead or alternatively continue to use the colon as separator instead (e.g. `class Foo : ParentClass, TraitA, TraitB`). IMHO it's a bit weird having a comma separated list of names after `impl` anyway (e.g. `class Foo impl TraitA, TraitB`) since this is so different than how that keyword is used in `impl X for Y`. With type parameters it would be even more different.

soc:
Agreed, my idea of picking something else ended up on `impls`, which is as close as possible to `impl` and it's kinda nice because one can argue that it implies that multiple impl things can follow, which is correct in this case.

I share your concern about commata – I'd see how it goes and adjust things when we have more experience reading this functionality.

soc:
Updated!

soc:
(You can squash the second commit, if you are fine with the PR.)

dinfuehr:
Btw i agree that `:` and `impl` doesn't look good, but imho also `extends` and `impls` don't really work together. One is abbreviated, the other not. `implements` would look better along with `extends` and be easier to distinguish from `impl`.

Personally I would even prefer the `:` syntax again (e.g. `class Foo: ParentClass, TraitA`). Since you wanted traits more explicit maybe prefixing with `trait` works as well: `class Foo: ParentClass, trait TraitA`

soc:
`implements` sounds fine to me, I don't care too much about the exact keyword.

The reason why I'm trying to get rid of `:` in this instance is that it introduces inconsistency around the meaning of `:`: 

Without it, `: B` means "the thing preceding it is of type B", but with this additional use in parent class syntax it grows an use that is completely different in both meaning and use.

I try to avoid this to make sure that "common" things only have one meaning, similar to preserving `[]` for types instead of e. g. using it for arrays _and_ types; or using `<>` as generics _and_ as binary operators.

soc:
@dinfuehr updated!

dora-parser/src/parser.rs
```diff
@@ -469,6 +472,19 @@ impl<'a> Parser<'a> {
         Ok(types)
     }
 
+    // Name is placeholder
+    fn parse_direct_trait_impls(&mut self) -> Result<Vec<Name>, ParseErrorAndPos> {
+        if self.token.is(TokenKind::Implements) {
+            self.advance_token()?;
+            let impls = self.parse_list_0(TokenKind::Comma, TokenKind::LBrace, false, |p| {
```
dinfuehr, line 479: Let's invoke another function here: parse_list_stop or so. parse_list and parse_list_stop would then use e.g. parse_list_inner internally.

dinfuehr:
Added the comment, when this is fixed we can merge this PR.

While I understand you have your own priorities, I wonder whether your time is well spent on this particular feature (and the annotation PR). Both features don't seem to be super high priority to me, simply because it's already possible to implement a trait for a class, a shortcut for that doesn't enable new stuff when writing Dora. Namespaces, lambdas, structs, etc. are all much more fundamental and would be much more useful right away.

soc:
> I wonder whether your time is well spent on this particular feature (and the annotation PR). Both features don't seem to be super high priority to me, simply because it's already possible to implement a trait for a class, a shortcut for that doesn't enable new stuff when writing Dora.

From my perspective, annotation thing are really fundamental and needed to be in place before namespacing, because it unhardcodes things in many places where we previously assumed they were hardcoded. From a safety perspective it's important that e. g. namespacing is able to properly tell e. g. `@open` coming from the std lib apart from an `@open` a user defined.

This PR is not that important (I think this is about as much language budget I'm ever willing to spend on something optional), but I feel this is as large as a project as I can reasonably manage without being stuck rebasing for half a year.

> Namespaces, lambdas, structs, etc. are all much more fundamental and would be much more useful right away.

I have already put some work/thoughts into namespaces, and have some syntax consistency/alignment plans for structs, but I feel that all three are much larger than what I can stomach – I don't feel the current development process scales beyond 1 person. (There is nothing wrong with that in general, but I think it limits the kind of contributions the project may get.)

dinfuehr:
The thing is I don't really see how user-defined annotations could be used at the moment. Assuming users can define annotations and annotate classes/methods/fields/etc. with them - how could a developer then use it? Wouldn't we need support for reflection (e.g. find all classes with a specific annotation)? Even if that all would work out, user-defined annotations seem to be of less priority than the other stuff I mentioned. Also not sure how we can define `@open` both in the stdlib and the user application without any support for namespaces.

> This PR is not that important (I think this is about as much language budget I'm ever willing to spend on something optional), but I feel this is as large as a project as I can reasonably manage without being stuck rebasing for half a year.
That's why I prefer smaller PRs, those will be much faster to merge and rebasing shouldn't be necessary as much. I can't review larger PRs on the side but have to reserve some time to properly think it through.

> I have already put some work/thoughts into namespaces, and have some syntax consistency/alignment plans for structs, but I feel that all three are much larger than what I can stomach
Those features can sound intimidating at first but once you split them up into smaller pieces of work, each step isn't so crazy anymore. If you are interested to work on one of them, we could discuss the implementation and how to split the work. Discussing this beforehand should also help to get reviews faster.

soc:
>  Also not sure how we can define @open both in the stdlib and the user application without any support for namespaces.

The idea why have annotations first, was that we can have tests to make sure namespacing works for them like any other element from the beginning.

PR #236, opened 2020-10-03
Upgrade to Rust nightly-2020-09-28 and update Cargo.lock

soc:


soc:
(2020-09-28 was the most recent  nightly with all components available.)

PR #241, opened 2021-02-05
Support sub-directories when munching standard library for inclusion

soc:


dora/build.rs
```diff
@@ -34,10 +35,15 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+    for f in WalkDir::new(SOURCE_DIR) {
         let f = f?;
 
-        if !f.file_type()?.is_file() {
+        if f.file_type().is_dir() {
+            create_dir_all(out_path.join(f.path()))?;
+            continue;
+        }
+
+        if f.file_type().is_symlink() {
```
dinfuehr, line 46: Why do we need to handle symlinks explicitly?

dinfuehr:
The problem that PR still has is that namespaces don't really work at all in the stdlib, which probably means we shouldn't make use of directories in the stdlib yet. Not sure we should merge this without that functionality. Supporting namespaces would also be useful for boots, which should also be embedded into the binary at some point.

dora/build.rs
soc: It's the preserved logic from before: !is_file --> is_dir || is_symlink

dinfuehr: Couldn't we just keep the current logic and just create the path before copying the file? That would have the advantage that we don't create directories without dora files.

soc: Yes, absolutely, thanks for the review, I changed it and it's much better that way!

soc:
My idea was to get some of the groundwork merged early, so that I can figure what needs to be done next to support dirs.

dora/build.rs
```diff
@@ -50,6 +51,7 @@ fn main() -> Result<(), Box<dyn Error>> {
             None => continue,
         }
 
+        create_dir_all(out_path)?;
```
dinfuehr, line 54: Shouldn't this be `out_path.join(f.path())` as well?

dinfuehr:
Great, let's merge this PR then after that comment is addressed.

soc:

dora/build.rs
```diff
@@ -34,10 +35,13 @@ fn main() -> Result<(), Box<dyn Error>> {
...
     writeln!(&mut stdlib, r#"["#,)?;
 
-    for f in fs::read_dir(SOURCE_DIR)? {
+    for f in WalkDir::new(SOURCE_DIR) {
         let f = f?;
 
-        if !f.file_type()?.is_file() {
+        if !f.file_type().is_file() {
+            if f.file_type().is_dir() {
+                create_dir_all(out_path.join(f.path()))?;
```
dinfuehr, line 43: Can we move this `create_dir_all` invocation, right before the copy call? Then we would only create directories with dora source files in them.

Cargo.lock
```diff
@@ -1,7 +1,5 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
```
dinfuehr, line 3: Can we install the walkdir crate without updating the crates in this PR? (we can do it in a separate PR)

dinfuehr:
Cool! Just as a note, I think this PR doesn't fully solve sub-directories in the standard library. I suspect there will still be problems with using/importing those files. Anyway, it's totally fine that this PR doesn't implement this all at once.

dora/build.rs
```diff
@@ -50,7 +51,10 @@ fn main() -> Result<(), Box<dyn Error>> {
             None => continue,
         }
 
-        copy(root_path.join(f.path()), out_path.join(f.path()))?;
+        if f.file_type().is_file() {
```
dinfuehr, line 54: Shouldn't this be always true here?

dinfuehr (approved):
Looks good with one comment

dora/build.rs
soc: Could also be a symlink.

dinfuehr: Hmm, but we already checked is_file() a few lines above, right? So this should always be true?

dinfuehr:
Thanks!

PR #237, opened 2020-10-17
Annotations part 2

soc:

PR #242, opened 2021-02-06
Annotations part 2

soc:
This is a rebase of https://github.com/dinfuehr/dora/pull/237 without the last two commits, which I'll work on further before creating a new PR.


soc:

dora-parser/src/parser.rs
```diff
@@ -523,16 +496,19 @@ impl<'a> Parser<'a> {
...
+    fn parse_class(
+        &mut self,
+        annotation_usages: AnnotationUsages,
+    ) -> Result<Class, ParseErrorAndPos> {
         let start = self.token.span.start();
-        let has_open = modifiers.contains(Modifier::Open);
-        let internal = modifiers.contains(Modifier::Internal);
-        let is_abstract = modifiers.contains(Modifier::Abstract);
-        let is_pub = modifiers.contains(Modifier::Pub);
+        let use_cannon = annotation_usages.iter().any(|annotation_usage| {
```
dinfuehr, line 508: Can you check whether we still need the `@cannon` annotation? I don't think we need it anymore.

dora/src/semck/clsdefck.rs
```diff
@@ -88,7 +88,14 @@ impl<'x> ClsDefCheck<'x> {
             AllowSelf::No,
         )
         .unwrap_or(SourceType::Error);
-        self.add_field(f.pos, f.name, ty, f.mutable, f.is_pub);
+        let is_pub = f.annotation_usages.contains(
```
dinfuehr, line 91: It seems we use such code a lot: Let's add a method `contains_pub` to AnnotationUsages.

dora/src/semck/extensiondefck.rs
```diff
@@ -127,19 +127,28 @@ impl<'x> ExtensionCheck<'x> {
     }
 
     fn visit_method(&mut self, f: &Arc<ast::Function>) {
-        if f.block.is_none() && !f.internal {
+        let internal = f.annotation_usages.contains(
```
dinfuehr, line 130: Same here.

dora/src/semck/extensiondefck.rs
```diff
@@ -232,7 +241,14 @@ impl<'x> ExtensionCheck<'x> {
             let method = self.vm.fcts.idx(method);
             let method = method.read();
 
-            if method.name == f.name && method.is_static == f.is_static {
+            let is_static = f.annotation_usages.contains(
```
dinfuehr, line 244: Same here.

dora/src/semck/extensiondefck.rs
```diff
@@ -256,7 +272,14 @@ impl<'x> ExtensionCheck<'x> {
             return true;
         }
 
-        let table = if f.is_static {
+        let is_static = f.annotation_usages.contains(
```
dinfuehr, line 275: Same here.

dora/src/semck/globaldef.rs
```diff
@@ -211,7 +211,14 @@ struct GlobalDef<'x> {
 
 impl<'x> visit::Visitor for GlobalDef<'x> {
     fn visit_namespace(&mut self, node: &Arc<ast::Namespace>) {
-        let id = NamespaceData::new(self.vm, self.namespace_id, node.name, node.is_pub);
+        let is_pub = node.annotation_usages.contains(
```
dinfuehr, line 214: Same here.

dora/src/semck/globaldef.rs
```diff
@@ -230,12 +237,19 @@ impl<'x> visit::Visitor for GlobalDef<'x> {
...
     fn visit_trait(&mut self, node: &Arc<ast::Trait>) {
         let id: TraitId = (self.vm.traits.len() as u32).into();
+        let annotations = &node.annotation_usages;
         let xtrait = TraitData {
             id,
             file_id: self.file_id,
             ast: node.clone(),
             namespace_id: self.namespace_id,
-            is_pub: node.is_pub,
+            is_pub: annotations.contains(
```
dinfuehr, line 246: Same here.

dora/src/semck/globaldef.rs
```diff
@@ -268,14 +282,21 @@ impl<'x> visit::Visitor for GlobalDef<'x> {
...
+            let annotations = &node.annotation_usages;
             let global = GlobalData {
                 id,
                 file_id: self.file_id,
                 ast: node.clone(),
                 namespace_id: self.namespace_id,
                 pos: node.pos,
                 name: node.name,
-                is_pub: node.is_pub,
+                is_pub: annotations.contains(
```
dinfuehr, line 293: Same here

dora/src/semck/globaldef.rs
```diff
@@ -359,8 +380,13 @@ impl<'x> visit::Visitor for GlobalDef<'x> {
                 internal: node.internal,
                 internal_resolved: false,
                 has_constructor: node.has_constructor,
-                is_pub: node.is_pub,
-
+                is_pub: annotations.contains(
```
dinfuehr, line 383: Same here

dora/src/semck/globaldef.rs
```diff
@@ -384,14 +410,21 @@ impl<'x> visit::Visitor for GlobalDef<'x> {
...
+            let annotations = &node.annotation_usages;
             let xconst = ConstData {
                 id,
                 file_id: self.file_id,
                 ast: node.clone(),
                 namespace_id: self.namespace_id,
                 pos: node.pos,
                 name: node.name,
-                is_pub: node.is_pub,
+                is_pub: annotations.contains(
```
dinfuehr, line 421: Same here

dora/src/semck/globaldef.rs
```diff
@@ -469,16 +470,29 @@ impl<'x> visit::Visitor for GlobalDef<'x> {
...
             let id: StructId = (structs.len() as u32).into();
+            let annotations = &node.annotation_usages;
             let mut xstruct = StructData {
                 id,
                 file_id: self.file_id,
                 ast: node.clone(),
                 namespace_id: self.namespace_id,
                 primitive_ty: None,
-                is_pub: node.is_pub,
+                is_pub: annotations.contains(
```
dinfuehr, line 480: Same here

dora/src/semck/globaldef.rs
```diff
@@ -469,16 +470,29 @@ impl<'x> visit::Visitor for GlobalDef<'x> {
...
+                    self.vm
+                        .annotations
+                        .idx(self.vm.known.annotations.pub_)
+                        .read()
+                        .name,
+                ),
                 pos: node.pos,
                 name: node.name,
-                internal: node.internal,
+                internal: annotations.contains(
```
dinfuehr, line 489: Same here

dora/src/semck/globaldef.rs
```diff
@@ -527,7 +548,13 @@ impl<'x> visit::Visitor for GlobalDef<'x> {
             name: node.name,
             type_params: Vec::new(),
             type_params2: TypeParamDefinition::new(),
-            is_pub: node.is_pub,
+            is_pub: annotations.contains(
```
dinfuehr, line 551: Same here

dora/src/semck/impldefck.rs
```diff
@@ -184,7 +191,14 @@ impl<'x> ImplCheck<'x> {
     }
 
     fn visit_method(&mut self, method: &Arc<ast::Function>) -> FctId {
-        if method.block.is_none() && !method.internal {
+        let internal = method.annotation_usages.contains(
```
dinfuehr, line 194: Same here.

dora/src/vm/known.rs
```diff
@@ -109,6 +110,22 @@ pub struct KnownTraits {
...
+    pub final_: AnnotationId,
+    pub internal: AnnotationId,
+    pub override_: AnnotationId,
+    pub open: AnnotationId,
+    pub pub_: AnnotationId,
+    pub static_: AnnotationId,
+
+    pub test: AnnotationId,
+
+    pub cannon: AnnotationId,
```
dinfuehr, line 125: Do we still need this one?

dinfuehr (approved):
Thanks, just a few nits...

dora/src/vm/known.rs
soc: I'm not sure? The annotation exists and is used 200 times.

dora-parser/src/parser.rs
soc: It's still used here:

```
    let code = match bc {
        CompilerName::Cannon => cannon::compile(vm, &fct, src, &type_params),
        CompilerName::Boots => boots::compile(vm, &fct, src, type_params),
    };
```

dora/src/semck/clsdefck.rs
soc: The problem is that we need `VM` to look up the name, so it would need to be done differently (or not in `AnnotationUsages`/dora-parser). What do you think?

soc: I guess one could move the methods to the `dora` module. I'll do that, if you don't have a different preference.

dora/src/vm/known.rs
soc: If we don't need it, I'm happy to remove things, but I think this should be a separate PR, not mixed with this one.

dora/src/semck/clsdefck.rs
dinfuehr: Sounds good.

dora/src/vm/known.rs
dinfuehr: Agreed, let's keep this. I thought that use_cannon above would only be a leftover...

dora-parser/src/parser.rs
dinfuehr: Right, there are still quite some uses of `@cannon`. Let's keep them here.

soc:
@dinfuehr Could you have a look? Your requested changes are in the last commit.

soc:

dora-parser/src/parser.rs
```diff
@@ -86,98 +86,70 @@ impl<'a> Parser<'a> {
     }
 
     fn parse_top_level_element(&mut self) -> Result<Elem, ParseErrorAndPos> {
-        let modifiers = self.parse_annotation_usages()?;
+        let annotation_usages = self.parse_annotation_usages()?;
 
         match self.token.kind {
             TokenKind::Fun => {
-                self.restrict_modifiers(
```
dinfuehr, line 93: When this is gone, is there still error reporting for illegal annotations?

dinfuehr (commented):
There is also a warning that needs to be fixed before merging.

I think I've mentioned it before but I believe we could merge code a lot faster (something you've complained about already) if we would split up such PRs. Here we could have easily introduced AnnotationUsages in addition to Modifers and only get rid of it when everything is in place.

dora-parser/src/lexer/position.rs
```diff
@@ -14,6 +14,10 @@ impl Position {
 
         Position { line: l, column: c }
     }
+
+    pub fn invalid() -> Position {
```
dinfuehr, line 18: Should we have this method at all? We are a bit circumventing around the type system and assertions? If we might not have a position we could use `Option<Position>` or if we prefer `Position` we could simply continue to use `Position::new(1, 1)`.

dora-parser/src/parser.rs
soc: I think this is reintroduced in annotations part 3, after the infrastructure is set in place in this part.

dora-parser/src/lexer/position.rs
soc: I don't care much either way, I was picking a way that would make it immediately obvious that the values are invalid, unlike `Position::new(1, 1)`, which is a valid row/column set that can appear in a real file.

soc:
@dinfuehr What's the status of this?

soc:

soc:

soc:
@dinfuehr could you have a look please?

dinfuehr:
Hey, I still don't get why you don't split up that PR. You've already created individual commits anyways, however with reviewing those don't really help. When you want to get your code in _sooner_, you have to create multiple PRs instead. I've mentioned this multiple times here already and TBH started to ignore this PR.

My general thought of annotations: I am quite skeptical in adding more stuff to them, simply because there is no way for users to access them at all. We can't even get a list of classes/functions with a specific annotation, nor can we check whether some class has some annotation listed. IMHO we should have that before we start to add properties/fields to annotations at all.

That said I am not even sure I want dynamic stuff like that at all in Dora since this probably makes AOT compilation tricky. And at least I've thought a few times about compiling Dora programs into an executable instead of running it in a VM.

soc:
Hey Dominik,

> You've already created individual commits anyways, however with reviewing those don't really help.

Sorry, I didn't know that this was the case, I would like to understand this better – what is GitHub doing/missing that makes a difference between reviewing the individual commits of a PR vs. reviewing one commit per PR?

> I am quite skeptical in adding more stuff to them, simply because there is no way for users to access them at all. We can't even get a list of classes/functions with a specific annotation, nor can we check whether some class has some annotation listed. IMHO we should have that before we start to add properties/fields to annotations at all.

For the current set of annotations, the compiler (and the human reader of the code, and possible doc generators) knowing about them is sufficient (e. g. `@deprecated`), so I didn't want (or need) to make a decision about user-accessibility of annotation in these PRs.

As there are multiple valid design approaches, ranging from doing runtime reflection (not a huge fan) to annotations using compiler hooks to act as macros (probably way too much power), this is something that needs a separate discussion on its own that I didn't want to mix into these changes that are largely focused on language and compiler hygiene.

I'm happy to brain storm about user-accessibility of annotations together, whenever it is convenient for you, though I believe I still have work left on annotations before this becomes acute.

> That said I am not even sure I want dynamic stuff like that at all in Dora since this probably makes AOT compilation tricky. And at least I've thought a few times about compiling Dora programs into an executable instead of running it in a VM.

I made sure to not introduce any mechanism that would make AOT harder as I have the same concerns for you. Everything that happens, happens at compile-time with these changes, as before. Anything going further would require more thought, and as mentioned, I wouldn't want to go down this road on my own.

Anyway, I will create individual PRs for each commit as soon as I manage to.

Thanks,

Simon

soc:
First commit is in https://github.com/dinfuehr/dora/pull/274.

soc:
Second commit, slightly adjusted, is https://github.com/dinfuehr/dora/pull/275.

PR #245, opened 2021-08-22
stdlib: implement String#startsWith and String#endsWith

soc:


soc:
std lib implementations of some string ops, they should probably be intrinsified in the long term, but having a basic implementation is needed anyway.

dinfuehr:
Thanks!

soc:
Next up is `contains` and `indexOf`, though I have trouble to decide which algorithm to pick.

(I think it doesn't matter much in the end, because the op should be intrinsified anyway, but finding something reasonably fast and easy to understand is challenging: https://www.researchgate.net/publication/332773245_Exact_String_Matching_Algorithms_Survey_Issues_and_Future_Research_Directions)

dinfuehr:
IMHO a naive algorithm would suffice for now but a more optimal implementation would definitely be great ;) I don't see how `contains` and `indexOf` would profit from intrinsification though.

soc:
On x86 at least you can use [PCMPESTRI](https://www.felixcloutier.com/x86/pcmpestri) if SSE4.2 is available. Not sure if ARM has something comparable ...

PR #246, opened 2021-08-23
stdlib: implement String#contains, String#indexOfFirst and Array::compare

soc:


dora/stdlib/String.dora
```diff
@@ -77,6 +77,58 @@
 
   @pub fun isEmpty(): Bool = self.size() == 0L;
 
+  @pub fun contains(val: String): Bool = self.indexOfFirst(val).isSome();
+
+  @pub fun indexOfFirst(val: String): Option[Int64] {
+    let m = val.size();
+    if m == 0L {
+      return Option[Int64]::Some(0L);
```
dinfuehr, line 85: `Some(0L)` should be good enough here.

dora/stdlib/String.dora
```diff
@@ -77,6 +77,58 @@
...
+    }
+
+    let n = self.size();
+    if m > n {
+      return Option[Int64]::None;
```
dinfuehr, line 90: `None` should be good enough here.

tests/array/array-compare.dora
```diff
@@ -0,0 +1,18 @@
+fun main() {
+  let a = Array[Int32](1, 2, 3, 3, 4, 5, 4, 5, 4, 3, 3);
+  let b = Array[Int32](1, 2, 3, 3, 4, 5, 4, 5, 4, 3, 4);
+  let c = Array[Int32](1, 2, 3, 3, 4, 5, 4, 5, 4, 3);
+  let d = Array[Int32](2, 3, 3, 4, 5, 4, 5, 4, 3, 4);
+  let e = Array[Int32]();
+
+  assert(Array[Int32]::compare(a, 0L, a, 0L, 11L));
+  assert(Array[Int32]::compare(a, 0L, a, 0L, 12L).not()); // desired?
```
dinfuehr, line 9: shouldn't this be fail because of out-of-bounds access?

soc: I haven't thought thoroughly whether `len` should be "exactly `len` elements" or "up to `len` elements", so I opted for the more "lenient" interpretation, with the plan to revisit this when I come around to write documentation for this method.

soc: I added  a todo documenting this.

dora/stdlib/String.dora
soc: Thanks, done!

dora/stdlib/String.dora
soc: Thanks, done!

dora/stdlib/Array.dora
```diff
@@ -100,6 +100,27 @@ impl[T: Equals] Array[T] {
 
     return false;
   }
+
+  // todo: decide whether to treat `len` as "exactly len elements" or "up to len elements".
+  // currently the (more lenient) latter interpretation is used.
+  @pub @static fun compare(a: Array[T], aPos: Int64, b: Array[T], bPos: Int64, len: Int64): Bool {
```
soc, line 106: new --^

soc:
@dinfuehr is this good to go, or do you want to have `compare` changed?

dinfuehr:
I would really prefer that we implement this as "exactly len elements", imho it's very uninituitive if `compare` would silently compare e.g. only 3 elements when the caller actually wants to compare say 10 elements. If the other behavior is needed, the caller can simply calculate the right length.

soc:
Ok, will do.

soc:
So instead of returning `false` we panic?

soc:
@dinfuehr Done!

dinfuehr:
Thanks!

Issue #240, opened 2020-10-27
Precedence of &


soc:
> So Ritchie maintained backwards compatibility forever and made the precedence order &&, &, ==, effectively adding a little bomb to C that goes off every time someone treats & as though it parses like +, in order to maintain backwards compatibility with a version of C that only a handful of people ever used.

[Eric Lippert: Hundred year mistakes](https://ericlippert.com/2020/02/27/hundred-year-mistakes/)

We should consider adjusting precedence based on this, some other languages that do it "correctly" are Swift, Go, Ruby and Python.

dinfuehr:
Hmm, isn't this what we already do?

dinfuehr:
Since I was already looking at the operator precedence, I shuffled it around a bit: https://github.com/dinfuehr/dora/commit/11c7a3b8a450c423fc7753072f30516abc8a739b

soc:
I should write some tests for this ... :-)

dinfuehr:
Sounds good.

soc:
Ok, great!

PR #247, opened 2021-08-23
Adjust precedence of &

soc:
Fixes https://github.com/dinfuehr/dora/issues/240.


soc:
@dinfuehr would have been better without squashing, as the test documented the before/after state ...

PR #249, opened 2021-08-24
Implement more methods for Option and Result

soc:


dora/stdlib/Result.dora
```diff
@@ -26,4 +40,80 @@ impl[V, E] Result[V, E] {
...
+  @pub fun or(alt: Result[V, E]): Result[V, E] {
+    if self.isOk() {
+      return self;
+    } else {
+      return alt;
+    }
+  }
+}
+
+impl[V: Equals, E: Equals] Result[V, E] {
```
dinfuehr, line 60: It doesn't look like `E: Equals` would be necessary here, right?

dinfuehr:
I have merged this PR anyways, please open another PR if my comment was correct.

dinfuehr:
Since the Zero trait is a bit special, maybe that's why the compiler doesn't find those methods... I will take a look.

dinfuehr:
I've looked at the bugs mentioned in the PR:

* One is related to Zero and the fact that there is no impl-Block for types implementing it. That's why cannon fails to produce machine code from the generated byte code. Somewhat related I think that `Int32::zero()` should also work but is an error atm. I am thinking about introducing special handling for auto-implemented traits here, I don't want to add impl-blocks to all types that implement the Zero trait since especially with Zero this seems tricky (can't just add it to all classes but also in the future tuples, structs, etc.). Also the number of auto-implemented traits should be quite small.

* The other one is related to comparing enums. I wanted to allow comparing of simple enums (integer comparison) but also unintentionally enabled this for all kinds of enums. I will have to at least limit this default comparison to those simple enums but a cleaner solution would maybe be a `Equals` annotation which can be used on simple enums and which makes the runtime generate the impl-block for equals as well.

dora/stdlib/Result.dora
soc: Oh, well ... it would be, if I hadn't implemented `equals` below incorrectly. :-D

I'll fix that, and then the `E : Equals` makes more sense probably.

soc:
- Agreed. I think we simply never needed this exact functionality until now, because for the array case we got away without it.

PR #254, opened 2021-08-28
Implement more methods on Array and Vec

soc:


dinfuehr:
Very cool and useful! Something we could discuss at a later point: Maybe those `toString` methods on Array and Vec could use a different name e.g. `inspect` or so like Ruby calls it. I would assume their output is extremely useful for developers but users probably don't really know what `Array` or `Vec` is.

soc:
Agreed! I had plans to write up a proposal regarding the distinction of `String` for debugging and `String` for user-facing information; I'll try speeding that up so we can discuss this.

soc:
@dinfuehr Perhaps it makes sense to activate the GitHub discussions feature, such that we don't clog up the issues tab with discussion topics?

dinfuehr:
I can activate it, I am just a bit worried that we get too many communication channels...

soc:
I'd say let's try it, we can always close it down again if it doesn't feel beneficial.

PR #255, opened 2021-08-28
stdlib: Extend Sortable and implement Array::sortStable with an insertion sort

soc:


dora/stdlib/Array.dora
```diff
@@ -72,6 +72,24 @@ impl[T: Zero] Array[T] {
   }
 }
 
+impl[T: Sortable] Array[T] {
+  // a stable in-place sort, currently implemented by insertion sort
+  @pub @static
+  fun sortStable(array: Array[T]) {
+    var i = 1L;
+    while i < array.size() {
```
dinfuehr, line 80: Code formatting seems a bit off..

dinfuehr:
Why do we have both Comparable and Sortable?

soc:
`Sortable` operations are the superset of `Identity`, while `Comparable` ops are the superset of `Equals`.

The idea is that `Sortable` implements the `totalOrder` predicate (IEEE754 §5.10) and `Comparable` implements the `comparison` predicate (IEEE754 §5.11).

dinfuehr:
Thanks, interesting. Out of curiosity and laziness: where does Rust suffer from that?

soc:
So this is my overview:

```
Java:     List.of(Double.NaN).contains(Double.NaN); // true
Scala:    List(Double.NaN).contains(Double.NaN)     // false
Rust:     &[0.0/0.0].contains(0.0/0.0)              // false
C#:       var list = new List<double>() { double.NaN };
          list.Contains(double.NaN);                // true
Haskell:  elem (0.0/0.0) [0.0/0.0]                  // false
```

The problem is not that languages decide on `true` or `false`, but that as a user one doesn't even get the choice of semantics.

In Rust's case they decided hat `contains` uses §5.11, which is fine by itself, but due to the subtyping relationship in Rust between `PartialEq` and `Eq`, Rust is unable to provide a different implementation i. e. according to §5.10. So you are stuck with it.

In Dora, we have `Identity` and `Equals`, which allows us to ask both ways, according to what we want to know or do.

E. g. "does this list include an element that is considered to be equal to the one I'm looking for" (`==` & `contains`) vs. "does this list include an element that is identical to the one I'm looking for" (`===` & `has`).

(I have started to think about ways of not needing to have two different methods through-out everything, but even at this state, Dora is far ahead in design quality and overall correctness compared to the "state of the art" on this topic.)

dinfuehr:
I will merge this PR and fix the formatting. Can you look into fixing sortAs as well?

soc:
Will do!

PR #259, opened 2021-09-02
More methods on number types and additional tests

soc:


soc:

Issue #257, opened 2021-08-30
Check whether raw-cpuid can be replaced by std macro


soc:
We currently use raw-cpuid crate for some CPU feature detection.

It turns out that Rust has https://doc.rust-lang.org/std/macro.is_x86_feature_detected.html since a while back, might make sense to use this instead of an external crate.

PR #260, opened 2021-09-03
x64: drop raw-cpuid dependency, use feature from std library

soc:
Fix https://github.com/dinfuehr/dora/issues/257.


dinfuehr:
Cool, thanks!

Issue #256, opened 2021-08-29
0.0/0.0 produces different results on x86 and Arm/RiscV


soc:
EDIT: Turns out this is not a bug in sorting, but caused by 0.0/0.0 producing different results on x86 and Arm/RiscV.

--- 

"negative" NaN and "positive" NaN are swapped:

```
// expected order
let nNaN = -(0.0/0.0);
let nInf = -1.0/0.0;
let nOne = -1.0;
let nZero = -0.0;
let pZero = 0.0;
let pOne = 1.0;
let pInf= 1.0/0.0;
let pNaN = 0.0/0.0;

// actual order
let x = Array[Float64](pOne, pInf, nInf, pNaN, nNaN, nZero, pZero, nOne);
Array[Float64]::sortStable(x);
assert(x(0L) === pNaN); // bug, should be at position 7
assert(x(1L) === nInf);
assert(x(2L) === nOne);
assert(x(3L) === nZero);
assert(x(4L) === pZero);
assert(x(5L) === pOne);
assert(x(6L) === pInf);
assert(x(7L) === nNaN); // bug, should be at position 0
```

spec:

> 5) If x and y are both NaNs, then totalOrder reflects a total ordering based on:
> i) negative sign orders below positive sign
> ii) signaling orders below quiet for +NaN, reverse for −NaN

dinfuehr:
An additional complication: NaNs (IIRC) seem to have a slightly different encoding on arm64, so sortsAs behaves differently on arm64 than on x64 with the current implementation. I had to disable the test for [sorting](https://github.com/dinfuehr/dora/commit/c8d0e1858cdf42066496b5703b486ff89ec80396) because of this.

soc:
Thanks, that's good to know!

dinfuehr:
I guess sortsAs then needs to be able to handle NaNs with some payload as well, right?

soc:
It should do that already, but it's possible that some architectures do not retain the exact payload.

(There is also the possibility that the quiet/signaling bit for NaNs is inverted. If you have the bitpattern it failed with then I can investigate further.)

dinfuehr:
Apparently positive and negative NaN are swapped on arm64 vs x64:

```
# arm64
pNaN = 7FF8000000000000
nNaN = FFF8000000000000

# x64
pNaN = FFF8000000000000
nNaN = 7FF8000000000000
```

[...]

PR #263, opened 2021-09-05
stdlib: more tests and documentation for Array.sortStable

soc:
Should probably close https://github.com/dinfuehr/dora/issues/256.

### DIFF
--- a/tests/array/array-sortStable.dora
+++ b/tests/array/array-sortStable.dora
@@ -1,45 +1,43 @@
-//= ignore
-import std::Assert;
-
 fun main() {
-  test1();
-  test2();
-  test3();
+  int32();
+  int64();
+  float64();
 }
 
-fun test1() {
+fun int32() {
   let x = Array[Int32](2, 3, 0, 4);
   Array[Int32]::sortStable(x);
   assert(x == Array[Int32](0, 2, 3, 4));
 }
 
-fun test2() {
-  let x = Array[Int32](5, 1, 4, 2, 8);
-  Array[Int32]::sortStable(x);
-  assert(x == Array[Int32](1, 2, 4, 5, 8));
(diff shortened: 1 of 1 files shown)


soc:
No idea why this fails ...

soc:
I dropped the string test part, things work now.

soc:

tests/array/array-sortStable.dora
```diff
@@ -1,45 +1,44 @@
...
-// this is an issue of the `sortsAs` implementation, not of the `sortStable` method.
-fun test3() {
+fun float64() {
+  // as x86 returns a negative NaN for 0.0/0.0, but Arm&RiscV return a positive NaN,
+  // we need to check first to make sure the NaN we are using has the desired sign.
+  let isX86 = (0.0/0.0).abs() !== 0.0/0.0;
+
   // expected order
-  let nNaN = -(0.0/0.0);
+  let nNaN = if isX86 { 0.0/0.0 } else { -(0.0/0.0) };
```
dinfuehr, line 25: Do we need `isX86` at all? Couldn't we build nNaN with `-(0.0/0.0).abs()` and pNaN with `(0.0/0.0).abs()`. IMHO this would be a bit clearer to me.

soc: I did it that way to make it a bit more obvious that this is related to platform-dependent behavior, but I can change it if you like.

dinfuehr: Yeah, I would prefer that. We have that comment of yours to explain why we do this...

soc: Ok, will do1

soc:

dinfuehr:
Thanks for pinging... Nice!

PR #265, opened 2021-09-07
DRAFT: stdlib: add Comparing and Sorting to replace Int-returning compareTo/sortAs methods

soc:


soc:
@dinfuehr Do you have an idea to turn `-1`/`0`/`1` into `Less`/`Equal`/`Greater` / `Before`/`Same`/`After` without introducing additional branches?

dinfuehr:
Enums without arguments should already be simple integers, so both enums should already use `0/1/2` as values. (We might want to change the encoding to `-1/0/1` but I guess that's initially not too important)

soc:
Will pick this up again at a later date.

PR #266, opened 2021-09-08
intrinsify `abs` on floats

soc:


dora/src/masm/x64.rs
```diff
@@ -1148,6 +1148,41 @@ impl MacroAssembler {
         }
     }
 
+    pub fn float_abs(&mut self, mode: MachineMode, dest: FReg, src: FReg) {
+        let (fst, snd) = if mode == MachineMode::Float32 {
+            (0, 0x7fffffff)
+        } else {
+            (0x7fffffff, -1)
```
soc, line 1155: I think the numbers aren't correct yet, so I didn't quite get how they were derived in `float_neg` either.

dora/src/masm/x64.rs
```diff
@@ -1148,6 +1148,41 @@ impl MacroAssembler {
...
+        }
+
+        if dest != src {
+            self.copy_freg(mode, dest, src);
+        }
+    }
+
     pub fn float_neg(&mut self, mode: MachineMode, dest: FReg, src: FReg) {
         let (fst, snd) = if mode == MachineMode::Float32 {
             (1i32 << 31, 0)
```
soc, line 1188: @dinfuehr Can you comment on how you arrived at these bits? I tried to implement it similarly for `float_abs`, but all combinations I try either produce the wrong result or crash the runtime altogether.

soc:

dora/src/cannon/codegen.rs
```diff
@@ -679,6 +679,21 @@ impl<'a> CannonCodeGen<'a> {
         self.emit_store_register(REG_RESULT.into(), dest);
     }
 
+    fn emit_abs_float(&mut self, dest: Register, src: Register) {
```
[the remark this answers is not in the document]
soc: shall I rename the other methods `emit_neg_float` etc. too?

soc:

dinfuehr:
Thanks!

soc:
👍

PR #275, opened 2021-11-28
Add Class::new and adjust Fct::new to allow easy replacement later

soc:


dora/src/vm/classes.rs
```diff
@@ -92,6 +92,68 @@ pub struct Class {
 }
 
 impl Class {
+    #[allow(dead_code)]
+    pub fn new(
```
dinfuehr, line 96: Looks like we should already use that method in `globaldef.rs` already? Then it wouldn't be dead code and gets test coverage.

dora/src/vm/functions.rs
```diff
@@ -118,6 +119,75 @@ impl Fct {
         }
     }
 
+    #[allow(dead_code)]
+    pub fn new_replacement(
```
dinfuehr, line 123: I would try to avoid adding a second method. I suspect this method was copy&pasted from above? I would suggest simply combining both methods. Should be possible if we would use e.g. `has_override: ast.has_override || annotations.contains(...)`? (Let's use a local variable for that `annotations.contains(...)` though. Then it should work for the old & new representation and after we've migrated e.g. `@override` fully, we can simply remove that `ast.has_override ||` part.

dora/src/vm/classes.rs
dinfuehr: I suspect this method was copy/pasted from there, right?

dora/src/vm/functions.rs
soc: good idea! will do.

soc: > I suspect this method was copy&pasted from above?

I think when I made these additions, the current method didn't exist.

dora/src/vm/classes.rs
soc: done.

dora/src/vm/functions.rs
soc: I reduced the scope of this commit, checking for annotations in `AnnotationUsages` requires having the annotations registered first.

dinfuehr:
Thanks!

soc:
Thanks, next PR coming!

soc:
@dinfuehr https://github.com/dinfuehr/dora/pull/276

PR #276, opened 2021-11-30
anno: visit annotations, register internal annotations 

soc:


dora/src/vm/annotations.rs
```diff
@@ -42,7 +42,8 @@ pub struct Annotation {
     pub name: Name,
     pub namespace_id: NamespaceId,
     pub ty: SourceType,
-    pub internal: Option<Modifier>,
+    pub internal: bool,
```
dinfuehr, line 45: Do we need both fields? Shouldn't `internal == internal_annotation.is_some()` hold?

dora/src/semck/stdlib.rs
```diff
@@ -239,6 +263,31 @@ fn internal_struct(
...
+fn internal_annotation(
+    vm: &mut VM,
+    namespace_id: NamespaceId,
+    name: &str,
+    internal_annotation: Modifier,
+) -> AnnotationId {
+    let name = vm.interner.intern(name);
+    let mut annotations = vm.annotations.lock();
+    let id: AnnotationId = annotations.len().into();
+    let annotation = Annotation {
```
dinfuehr, line 275: I am a bit confused about this code. In this file we usually only find e.g. classes or functions by name to get e.g. their IDs or set some specific properties. We generally create such data structures in src/semck/globaldef.rs.

dora/src/semck/stdlib.rs
```diff
@@ -239,6 +263,31 @@ fn internal_struct(
...
+        id,
+        name,
+        file_id: FileId::from(0),
+        pos: Position::invalid(),
```
dinfuehr, line 279: I know I've used `Position::new(1, 1)` a lot, but `Position::invalid()` violates the asserts in the `Position::new` constructor. If we really need something like invalid positions, imho we should leverage the type system and use `Option<Position>`.

soc: I'm fine with `Position::new(1, 1)`, the idea with an illegal position was that it was immediately obvious that this things was not correctly initialized. Making it an `Option` burdens the callers with a possibility they should never have to deal with (because the invalid state is transient, the next phase is supposed to fill in the correct details).

dora/src/semck/stdlib.rs
soc: I think the visitor in globaldef needs to be extended at a later stage to handle annotations that aren't "special", but at the same time we need to deal with these internal annotations early on – do you have a suggestion how it would work better?

(We could extend the visitor now and call `parse_file` with explicit arguments (or something along the lines of `parse_bundled_stdlib` ... do you have preferences?))

dora/src/vm/annotations.rs
soc: I think it was intended to reduce confusion between an internal annotation and *the* internal annotation.

I removed it.

soc:
@dinfuehr Could you have a look? Annotations are now visited, registering internal ones now has to be done after `globaldef::check` as they now do only lookup + set `internal_annotation` – this will cause trouble down the line when we migrate how the modifier fields are initialized, but ... baby steps, I guess.

dinfuehr:
Thanks!

PR #277, opened 2021-12-11
anno: add helper functions to recognize internal annotations

soc:


soc:

soc:
I can't reproduce the ASAN failure locally. :-/

soc:
Error is:

```
#==== STDERR
==86350==LeakSanitizer has encountered a fatal error.
==86350==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1
==86350==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)
```

soc:

dora/src/vm/annotations.rs
```diff
@@ -72,4 +72,63 @@ impl Annotation {
...
         self.internal_annotation.is_some()
     }
+
+    pub fn is_abstract(annotation_usages: &AnnotationUsages, vm: &VM) -> bool {
+        let name = vm
+            .annotations
+            .idx(vm.known.annotations.abstract_)
+            .read()
+            .name;
+        annotation_usages.contains(name)
```
dinfuehr, line 82: Can you please remind me: Why do we use the `name` here? Is this an intermediate step or will this stay this way?

I guess ideally we would resolve annotations immediately in a separate pass right at the beginning of `semck`, which would allow us to just use the id afterwards.

soc: > Is this an intermediate step or will this stay this way?

It's an immediate step, and a very coarse approximation of the desired behavior, because at this stage it would (incorrectly) accept any annotation called e. g. `abstract`, not only the one in `stdlib`.

> I guess ideally we would resolve annotations immediately in a separate pass right at the beginning of semck, which would allow us to just use the id afterwards.

Agreed. In the end an annotation is just another symbol that is brought into scope with a (by-default) import, so normal name resolution should run its course and deal with most of the questions.

dinfuehr: Cool, sounds great to me!

For the eventual follow-up work: I guess we need to initialize all these `is_abstract`/etc. flags of classes/functions/etc. not in globaldef.rs but later when we fully set them up. At a first glance, this doesn't look too tricky at least for builtin annotations.

dinfuehr:
Thanks!

dora/src/vm/annotations.rs
soc: Agreed! In a next commit I put the annotations in its own "stdlib-boot" directory and typecheck them there, so they should be fully present before the rest of the stdlib makes use of them.

dinfuehr: I think I would prefer if we wouldn't need a special directory for them - so far this looks feasible.

dinfuehr:
BTW I think you could add an annotation `@Test` and migrate that testing mechanism to that annotation.

dora/src/vm/annotations.rs
soc: Could you expand on this, i. e. where and when things should happen?

dinfuehr: Sure, I think we could simply create annotations in [globaldef.rs](https://github.com/dinfuehr/dora/blob/master/dora/src/semck/globaldef.rs) - like we do already. Right now we also initialize all those is_abstract, etc. flags in that same pass for classes/structs/etc. However this doesn't work anymore since we now need to set those flags after running `stdlib::resolve_internal_annotations` in [semck.rs](https://github.com/dinfuehr/dora/blob/master/dora/src/semck.rs#L56).

We could initialize those flags with `false` in globaldef.rs and properly set & check them in a pass afterwards (e.g. [clsdefck.rs](https://github.com/dinfuehr/dora/blob/master/dora/src/semck/clsdefck.rs) for classes). Alternatively we could add a new pass that does that work for classes, structs, etc (e.g. annotationck.rs). Not sure what's better atm.

Maybe we discover a reason why this doesn't work but so far it seems feasible.

soc: Ah ok, understand the idea. My reasoning for bootstrapping the annotations first was to avoid adding another pass or having some intermediate state where things aren't initialized properly. Will think about this!

Issue #140, opened 2020-01-26
Implement rounding of floating point numbers


soc:
This is probably a not-so-easy task given the amount of possible rounding modes and the poor support for rounding on x86, see https://gcc.godbolt.org/z/zEvearfc8.

For completeness, here is a list of signatures for possible rounding methods for `Float64`:

```
                                        C                x86                   arm64
fun roundAwayFromZero(): Float64;
fun roundToZero(): Float64;        // trunc             RC 0b11       frintz ("toward zero")
fun roundUp(): Float64;            // ceil              RC 0b10       frintp ("toward plus infinity")
fun roundDown(): Float64;          // floor             RC 0b01       frintm ("toward minus infinity")

fun roundHalfEven(): Float64;      // rint/nearbyint¹   RC 0b00       frintn ("to nearest with ties to even")
fun roundHalfOdd(): Float64;
fun roundHalfUp(): Float64;        // round¹            <magic>       frinta ("to nearest with ties to away")
fun roundHalfDown(): Float64;
```

working on this issue does not require implementing all of them, even a single one is fine!

---

¹ with `FE_TONEAREST`

soc:
```
roundToZero
roundUp
roundDown
roundHalfEven
```

have been implemented, rest should go on separate issues if there is demand.

Issue #267, opened 2021-09-08
Avoid SSE-coded instructions, prefer VEX-coded instructions if AVX is available


soc:
This is due the high cost of mixing SSE and AVX instructions, see https://john-h-k.github.io/VexTransitionPenalties.html.

PR #279, opened 2022-01-04
stdlib: implement first rounding methods for `Float32` and `Float64`

soc:
See https://github.com/dinfuehr/dora/issues/140.

### DIFF
--- a/dora-asm/src/x64.rs
+++ b/dora-asm/src/x64.rs
@@ -571,6 +571,26 @@ impl AssemblerX64 {
         self.emit_modrm_sse_registers(dest, src);
     }
 
+    pub fn roundss_ri(&mut self, reg: XmmRegister, imm: Immediate) {
+        self.emit_u8(0x66);
+        self.emit_rex_sse_modrm_optional(reg, reg);
+        self.emit_u8(0x0f);
+        self.emit_u8(0x3a);
+        self.emit_u8(0x0a);
+        self.emit_modrm_sse_registers(reg, reg);
+        self.emit_u8(imm.uint8());
+    }
+
+    pub fn roundsd_ri(&mut self, reg: XmmRegister, imm: Immediate) {
+        self.emit_u8(0x66);
+        self.emit_rex_sse_modrm_optional(reg, reg);
+        self.emit_u8(0x0f);
+        self.emit_u8(0x3a);
+        self.emit_u8(0x0b);
+        self.emit_modrm_sse_registers(reg, reg);
+        self.emit_u8(imm.uint8());
+    }
+
     pub fn sqrtss_rr(&mut self, dest: XmmRegister, src: XmmRegister) {
         self.emit_u8(0xf3);
         self.emit_rex_sse_modrm_optional(dest, src);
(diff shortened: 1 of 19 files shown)


soc:

dora/stdlib/Float32.dora
```diff
@@ -33,6 +33,7 @@ impl Float32 {
   @pub @internal fun unaryPlus(): Float32;
   @pub @internal fun unaryMinus(): Float32;
   @pub @internal fun abs(): Float32;
+  @pub @internal fun roundHalfEven(): Float32;
```
soc, line 36: I'm currently planning on having a method for each of the ~four rounding modes.

Alternative design is having one `fun round(RoundingMode)` method, but I first like to wait and see in how many additional places we would use such an enum.

dora-asm/src/x64.rs
```diff
@@ -571,6 +571,26 @@ impl AssemblerX64 {
         self.emit_modrm_sse_registers(dest, src);
     }
 
+    pub fn roundss_ri(&mut self, reg: XmmRegister, imm: Immediate) {
+        self.emit_u8(0x66);
+        self.emit_rex_sse_modrm_optional(reg, reg);
+        self.emit_u8(0x0f);
+        self.emit_u8(0x3a);
+        self.emit_u8(0x0a);
+        self.emit_modrm(0b11, reg.low_bits(), reg.low_bits());
```
dinfuehr, line 580: You can use `emit_modrm_sse_registers` here instead.

dinfuehr:
Why would we want to add e.g AVX encoding as well? I suppose advantages (if there are any) are miniscule at best but adds one more configuration to Dora.

dora-asm/src/x64.rs
soc: thanks, will do

soc:
Hi @dinfuehr, this is due to https://github.com/dinfuehr/dora/issues/267.

dinfuehr:
We can add that later but let's not add all these optimizations which atm just complicate the code but do not bring any real speedup.

soc:
Ok!

dora-asm/src/x64.rs
soc: done!

soc:
@dinfuehr I think one of your refactorings broke the linter.

dinfuehr:
Thanks, should be fixed now.

soc:
@dinfuehr all green!

dora/src/masm/x64.rs
```diff
@@ -1295,6 +1295,62 @@ impl MacroAssembler {
         }
     }
 
+    pub fn float_round_tozero(&mut self, mode: MachineMode, dest: FReg, src: FReg) {
+        if !has_round() {
+            panic!("unsupported operation")
```
dinfuehr, line 1300: What should we do when those instructions aren't supported? Do we need a workaround? From what I could find online in a minute or so, it seems we can safely assume those instructions exist since they seem old enough. In this case I would just let the CPU trap on these instructions and get rid of `has_round()` completely. Wdyt?

soc: My intention was to leave the door open for an asm fallback implementation, but I think that requiring a CPU from 2007 or later is reasonable. I'll drop that code.

soc:
@dinfuehr I think this should be good to go, the CI failure seems to be unrelated.

soc:
@dinfuehr could you he a look?

soc:
@dinfuehr ping?

PR #281, opened 2022-01-05
anno: typecheck annotations first, switch from boolean flags in the AST to annotation usages

soc:
WIP, will split up the two topics as soon as things work.


soc:
@dinfuehr didn't see https://github.com/dinfuehr/dora/pull/277#discussion_r778619754, will rework this when I know more

PR #285, opened 2022-02-02
anno: introduce annotationck phase that initializes modifiers

soc:
This shows that the approach of initializing the modifiers in the individual *ck-phases might work if it wasn't for the `X_accessible_from` access-checks:

If the `X` is later than the phase we already checked, then it fails because `is_public` is `false`, because the modifiers haven't been initialized yet. So e. g. class → struct works, but struct → class does not.

### DIFF
--- a/dora/src/language/impldefck.rs
+++ b/dora/src/language/impldefck.rs
@@ -195,13 +195,7 @@ impl<'x> ImplCheck<'x> {
 
         let parent = FctParent::Impl(self.impl_id);
 
-        let fct = FctDefinition::new(
-            self.sa,
-            self.file_id.into(),
-            self.namespace_id,
-            method,
-            parent,
-        );
+        let fct = FctDefinition::new(self.file_id.into(), self.namespace_id, method, parent);
         self.sa.add_fct(fct)
     }
 }
--- a/dora/src/vm/annotations.rs
+++ b/dora/src/vm/annotations.rs
@@ -1,3 +1,4 @@
+use crate::language::error::msg::SemError;
 use crate::language::ty::SourceType;
 use crate::utils::GrowableVec;
 use crate::vm::{FileId, NamespaceId, TypeParam, VM};
@@ -130,5 +131,27 @@ impl AnnotationDefinition {
             .read()
             .name;
         annotation_usages.contains(name)
+    }
+
+    pub fn reject_modifiers(
+        vm: &VM,
+        file_id: FileId,
+        modifiers: &AnnotationUsages,
+        rejects: &[AnnotationDefinitionId],
+    ) {
+        for reject in rejects.iter() {
+            let reject = vm.annotations.idx(*reject).read().name;
+            for modifier in modifiers.iter() {
+                if modifier.name == reject {
+                    vm.diag.lock().report(
+                        file_id,
+                        modifier.pos,
+                        SemError::MisplacedAnnotation(
+                            vm.interner.str(modifier.name).as_str().into(),
+                        ),
+                    );
+                }
+            }
+        }
     }
 }
--- a/dora/src/vm/namespaces.rs
+++ b/dora/src/vm/namespaces.rs
@@ -5,7 +5,7 @@ use dora_parser::ast;
 use dora_parser::interner::Name;
 
 use crate::language::sym::SymTable;
-use crate::vm::{FileId, VM};
+use crate::vm::{AnnotationDefinition, FileId, SemAnalysis, VM};
 
 #[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
 pub struct NamespaceId(pub usize);
@@ -74,6 +74,29 @@ impl NamespaceData {
             is_pub: ast.is_pub,
             parents,
             depth,
+        }
+    }
+
+    pub fn init_modifiers(&mut self, sa: &SemAnalysis) {
+        #[allow(unused_variables)]
+        if let Some(ast) = &self.ast {
+            let annotation_usages = &ast::AnnotationUsages::new();
+            self.is_pub = AnnotationDefinition::is_pub(annotation_usages, sa);
+            AnnotationDefinition::reject_modifiers(
+                sa,
+                self.file_id,
+                annotation_usages,
+                &[
+                    sa.known.annotations.abstract_,
+                    sa.known.annotations.final_,
+                    sa.known.annotations.internal,
+                    sa.known.annotations.open,
+                    sa.known.annotations.optimize_immediately,
+                    sa.known.annotations.override_,
+                    sa.known.annotations.static_,
+                    sa.known.annotations.test,
+                ],
+            );
         }
     }
 
--- a/dora/src/vm/namespaces.rs
+++ b/dora/src/vm/namespaces.rs
@@ -1,11 +1,11 @@
 use parking_lot::RwLock;
 use std::sync::Arc;
 
-use crate::language::sym::SymTable;
-use crate::vm::VM;
-
-use dora_parser::ast::Namespace;
+use dora_parser::ast;
 use dora_parser::interner::Name;
+
+use crate::language::sym::SymTable;
+use crate::vm::{FileId, VM};
 
 #[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
 pub struct NamespaceId(pub usize);
@@ -25,7 +25,8 @@ impl From<usize> for NamespaceId {
 #[derive(Debug)]
 pub struct NamespaceData {
     pub id: NamespaceId,
-    pub ast: Option<Arc<Namespace>>,
+    pub file_id: FileId,
+    pub ast: Option<Arc<ast::Namespace>>,
     pub parent_namespace_id: Option<NamespaceId>,
     pub name: Option<Name>,
     pub table: Arc<RwLock<SymTable>>,
@@ -38,6 +39,7 @@ impl NamespaceData {
     pub fn predefined(id: NamespaceId, name: Option<Name>) -> NamespaceData {
         NamespaceData {
             id,
+            file_id: FileId(0),
             ast: None,
             parent_namespace_id: None,
             name,
@@ -48,7 +50,12 @@ impl NamespaceData {
         }
     }
 
-    pub fn new(vm: &mut VM, parent_id: NamespaceId, ast: &Arc<Namespace>) -> NamespaceData {
+    pub fn new(
+        vm: &mut VM,
+        file_id: FileId,
+        parent_id: NamespaceId,
+        ast: &Arc<ast::Namespace>,
+    ) -> NamespaceData {
         let id: NamespaceId = vm.namespaces.len().into();
 
         let parent = &vm.namespaces[parent_id.to_usize()].read();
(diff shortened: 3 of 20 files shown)


soc:
@dinfuehr ideas?

soc:
@dinfuehr if there aren't any ideas how to deal with this, I'll try the way of moving the bits and pieces that do outline parsing in *ck phases to globaldef, and add an annotation phase after it.

dinfuehr:
Hmm, not sure I understand what you mean with that?

soc:
> We already create the "empty" annotation (like class/function/etc.) in globaldef and then right afterwards resolve builtin annotations. Let's pick Class::is_pub as an example. Right now we set that field during globaldef, IIRC what I meant was that we always set it to false in GlobalDef and then during clsdefck iterate the annotations on the class and if we find @pub set it to true here.

So this hits the issue mentioned above. So instead of doing this, the idea is to have an annotation-phase that initializes the modifiers rather early (~after outline parsing and import processing).

But to do this, it requires moving the creation of the constructors, methods, fields etc. from classdeck, structdefck to globaldefck (where the majority is already created).

dinfuehr:
I would rather initialize annotations for classes/structs/functions in a separate phase right after `globaldef`. We can't define e.g. functions/methods in `globaldef` anyways since not all types exist at that point. That design with `globaldef` exists for a reason (I didn't start with this design from the start but had to switch to it). I suspect there are more problems/dependencies like this.

soc:
> I would rather initialize annotations for classes/structs/functions in a separate phase right after globaldef

Yes, that's what I had in mind. globaldef is doing outline parsing (parsing declarations/definitions without looking at their bodies), but there are bits and pieces missing from it.

What I'm proposing is completing the missing parts of the outline phase and initializing the annotations in a phase after it.

I was thinking of doing it after import resolution, such that it is easier in the next step to switch from a name-based approach ("any annotation `@pub` is considered to be a modifier") to working on the resolved symbols ("only `@pub` annotations referring to `std::pub` set modifiers").

> We can't define e.g. functions/methods in globaldef anyways since not all types exist at that point.

Can you expand on that?

> I didn't start with this design from the start but had to switch to it

So you started out with doing all outline parsing in globaldef, and moved things to later phases – or you did all things in later phases first and then moved things to globaldef?

dinfuehr:
Ah right, fields/methods can have annotations as well. Not sure though, we want to move creation of those into globaldef (it's globaldef and not globaldefck btw - globaldefck is just for global variables. Naming is definitely unfortunate atm). When we create fields/methods all annotations should already be created and all internal ones should be known to the compiler already, so we should just be able to create fields/etc. directly with initialized `is_pub`/etc.

dinfuehr:
I like that globaldef only defines top-level elements. I would try to avoid adding fields/methods in `globaldef`.

I think (without actually looking it up) initially I started by roughly creating functions with argument and return types first (forward declarations were never necessary and Dora only had builtin types). And a second pass that then checked the function body.

When adding classes I just added an additional pass that created classes with field and method types just before the first function pass. I think I only realized a bit later that classes needed forward declarations that way (e.g `class Foo(x: Bar) class Bar(x: Foo)` didn't work afair). I think this and other new stuff forced me to introduce `globaldef`.

Type parameters are needed for `read_type` (we assume that it only returns valid types) to work: in `class Foo[T: Bar]` we need the trait `Bar` to be created already. Even more tricky if traits can be generic as well (that's not supported atm), something like `class Foo[T: Bar[Foo[Int]]]` might be legal.

soc:
> it's globaldef and not globaldefck

Apologies, I always meant to refer to the former.

> Not sure though, we want to move creation of those into globaldef

Ok, then I'll move only those created in globaldef required to the new, earlier phase.

> I like that globaldef only defines top-level elements.

I believe that having an outline phase (i. e. a phase that enters definitions, regardless of the level) would be more reliable and would avoid having to special-case things (this feature is likely not the last time this problem will come up). Though given the workaround, this is hopefully not something that needs to be addressed before this PR.

dinfuehr:
> > it's globaldef and not globaldefck
> 
> Apologies, I always meant to refer to the former.

No need to apologize. I just wanted to make sure we are talking about the same thing. Naming is definitely problematic for globaldef and globaldefck.

soc:
Ok, the earlier phase looks good except that I can't set modifiers for namespaces, because I can't get mutable values from `namespaces: Vec<NamespaceData>` in VM.

Is this intentional? (What's the general logic whether something is held as `GrowableVec<RwLock<X>>`, `GrowableVec<X>`, `Vec<RwLock<X>` or `Vec<X>`?)

soc:
@dinfuehr see https://github.com/dinfuehr/dora/pull/287, this would be a prerequisite to further work here.

dinfuehr:
> Ok, the earlier phase looks good except that I can't set modifiers for namespaces, because I can't get mutable values from `namespaces: Vec<NamespaceData>` in `VM`.
> 
> Is this intentional? (What's the general logic whether something is held as `GrowableVec<RwLock<X>>`, `GrowableVec<X>`, `Vec<RwLock<X>` or `Vec<X>`?)

I tried to use the simplest type possible, since NamespaceData wasn't mutable so far, `RwLock` wasn't necessary.

soc:
Ah, thanks understood.

soc:
@dinfuehr Could you have a look? Next step would be to switch the comparisons of names to a comparison of the actual thing.

Not sure what's the best way to replace the modifier tests that previously happened in parser.rs.

soc:
@dinfuehr ping?

soc:
CI failure is due to: `Error: Version 2.6.x not found`

dinfuehr:
Hey, I am not going to land this PR as-is. It's way too big to properly review it.

soc:
@dinfuehr I pulled out the addition of supporting methods into its own commit, please have a look.

dora/src/vm/annotations.rs
```diff
@@ -131,4 +132,25 @@ impl AnnotationDefinition {
             .name;
         annotation_usages.contains(name)
     }
+
+    pub fn reject_modifiers(
```
dinfuehr, line 136: I think we should define what modifiers we allow?

dora/src/vm/annotations.rs
```diff
@@ -131,4 +132,25 @@ impl AnnotationDefinition {
...
+        vm: &VM,
+        modifiers: &AnnotationUsages,
+        rejects: &[AnnotationDefinitionId],
+    ) {
+        for reject in rejects.iter() {
+            let reject = vm.annotations.idx(*reject).read().name;
+            for modifier in modifiers.iter() {
+                if modifier.name == reject {
+                    vm.diag.lock().report(
+                        FileId(0),
```
dinfuehr, line 146: We should definitely report the right filename and position here.

dora/src/vm/namespaces.rs
```diff
@@ -69,6 +69,28 @@ impl NamespaceData {
         }
     }
 
+    pub fn init_modifiers(&mut self, sa: &SemAnalysis) {
+        #[allow(unused_variables)]
```
dinfuehr, line 73: Why?

dora/src/vm/traits.rs
```diff
@@ -46,6 +46,48 @@ pub struct TraitDefinition {
...
+            methods: Vec::new(),
+            instance_names: HashMap::new(),
+            static_names: HashMap::new(),
+            vtables: RwLock::new(HashMap::new()),
+        }
+    }
+
+    pub fn init_modifiers(&mut self, sa: &SemAnalysis) {
+        let annotation_usages = &ast::AnnotationUsages::new();
+        self.is_pub = AnnotationDefinition::is_pub(annotation_usages, sa);
```
dinfuehr, line 74: It seems we always create empty AnnotationUsages and then directly call `is_pub` on it? Am I right and if so, why are we doing this here?

soc: I removed the changes that added `AnnotationUsages` to the AST.

dora/src/vm/annotations.rs
soc: Will do.

dora/src/vm/namespaces.rs
soc: The AST hasn't `AnnotationUsages` yet.

dora/src/vm/annotations.rs
soc: I think we talked about this already. With user-definable annotations, the set of all annotations isn't fixed, so flipping the check would require more awkward logic, not less, because we would have do to additional checks with regard to internal vs. user-defined.

soc: Tbh, I don't really care anymore, this isn't code that will be with us for a lot further.

dora/src/vm/annotations.rs
soc: It uses the `file_id` from the annotated element now, position should be fine, or do you want that to be done differently?

soc: (I added a preceding commit that adds `FileId` to `NamespaceData`, which did not have it before.)

soc:
@dinfuehr I rebased the commits on master such that the fixed CI could run again, can you have a look?

dora/src/vm/namespaces.rs
```diff
@@ -48,7 +50,12 @@ impl NamespaceData {
         }
     }
 
-    pub fn new(vm: &mut VM, parent_id: NamespaceId, ast: &Arc<Namespace>) -> NamespaceData {
+    pub fn new(
+        vm: &mut VM,
+        file_id: FileId,
```
dinfuehr, line 55: Not sure this makes sense atm, namespaces can span multiple files (a whole directory) at the moment, so there isn't a single file a namespace is defined in.

dora/src/language/impldefck.rs
```diff
@@ -195,13 +195,7 @@ impl<'x> ImplCheck<'x> {
...
         let parent = FctParent::Impl(self.impl_id);
 
-        let fct = FctDefinition::new(
-            self.sa,
-            self.file_id.into(),
-            self.namespace_id,
-            method,
-            parent,
-        );
+        let fct = FctDefinition::new(self.file_id.into(), self.namespace_id, method, parent);
```
dinfuehr, line 198: Why not open a separate PR for this refactoring? That we could land immediately and make this PR a bit smaller.

dora/src/vm/annotations.rs
dinfuehr: I don't see how this would be more awkward, the logic would just be `is_user_defined() || is_allowed_internal_annotation()`. Also we probably don't allow all user-defined annotations in all locations in the future (e.g. an annotion which is only allowed on methods).

dora/src/vm/namespaces.rs
dinfuehr: So AFAIU these init_modifiers aren't invoked atm, right? Also the code in there isn't really used atm. I don't think it makes sense to merge this PR as-is.

Why don't you first add a method `parse_annotation_usages` to the parser first that returns `AnnotationUsages`? AFAICS this method doesn't exist yet either. You could return both `Modifiers` and `AnnotationUsages` from that method to not break already existing code. Then switch element-by-element from `Modifiers` to `AnnotationUsages`.

PR #287, opened 2022-02-06
change `Vec<NamespaceData>` → `Vec<RwLock<NamespaceData>>` to allow mutation of namespaces later on

soc:
This is needed to initialize modifiers later on.


soc:
Thanks!

